PR #14, opened 2018-11-05
Observation endpoints

lboeman:
Added skeleton endpoints with some demo responses and TODO comments where things will have to be  updated.

### DIFF
--- a/sfa_api/observations.py
+++ b/sfa_api/observations.py
@@ -17,8 +17,9 @@ blp = Blueprint(
 )
 
 
-# TODO: Replace the demo content in these classes.
+# TODO: Replace the static demo content in these classes.
 class Site(object):
+    uuid = '123e4567-e89b-12d3-a456-426655440001'
     name = 'Ashland OR'
     resolution = '1 min'
     latitude = 42.19
@@ -33,7 +34,22 @@ class Site(object):
 
 class Observation(object):
     uuid = '123e4567-e89b-12d3-a456-426655440000'
-    site = Site()
+    variable = 'ghi'
+    site = '123e4567-e89b-12d3-a456-426655440001'
+
+
+class ObservationValue(object):
+    timestamp = '2018-11-05T18:19:33+00:00'
+    value = 35
+
+
+@spec.define_schema('ObservationValue')
+class ObservationValueSchema(ma.Schema):
+    class Meta:
+        strict = True
+        ordered = True
+    timestamp = ma.fields.DateTime()
+    value = ma.fields.Float()
 
 
 @spec.define_schema('Site')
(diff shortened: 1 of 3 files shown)


sfa_api/observations.py
```diff
@@ -58,17 +98,231 @@ def get(self, *args):
...
+        - $ref: '#/components/parameters/uuid'
+        requestBody:
+          description: JSON respresentation of an observation.
+          required: True
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Observation'
```
alorenzo175: probably supposed to by ObservationValue?

sfa_api/observations.py
```diff
@@ -27,28 +74,21 @@ class ObservationSchema(ma.Schema):
     class Meta:
         strict = True
         ordered = True
-    uuid = ma.fields.UUID(description='UUID response')
-    site = ma.fields.Nested(SiteSchema)
+    uuid = ma.fields.UUID()
+    variable = ma.fields.String()
+    site = ma.fields.UUID()
```
alorenzo175: Not very Restful maybe, but what about having ``site_id`` be the uuid and a ``site`` dict with the actual site info? All the use cases I see for the observation metadata also require some keys from the site metadata

lboeman: I think that's fair, it avoids having to make multiple http requests in many cases.  We'll just have to document that it loads the site metadata for convenience, and that only a uuid should be included in a post request. 
We also might need to add human-readable name-field to these. 
On a similar note, we will probably need to list these per site, where it can be safely assumed that an api user/application already has site data. Should we still have one endpoint to list all obs/data points with just uuids  and type and name on an /observations/metadata endpoint or should they all be included on /observations?

alorenzo175: A name field might be something like {site} {variable} but that would cause issues if the site name changed. If the full site is included, I think that + variable is sufficient.

I think we should just stick with /observations (which includes the full site dict) for now.

sfa_api/observations.py
```diff
@@ -58,17 +100,230 @@ def get(self, *args):
...
+        description: Create a new Observation by posting metadata.
+        requestBody:
+          description: JSON respresentation of an observation.
+          required: True
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  site:
```
alorenzo175: site_id

sfa_api/observations.py
```diff
@@ -58,17 +100,230 @@ def get(self, *args):
...
+          - $ref: '#/components/parameters/uuid'
+        requestBody:
+          description: JSON representation of an observation's metadata.
+          required: True
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  site:
```
alorenzo175: site_id

lboeman:
No there isn't. I was attempting to find a way to use the same marshmallow schema and define both a GET and a POST schema for openapi using different subsets of its fields, but I don't think that's possible.

PR #18, opened 2018-11-07
add forecast endpoints

lboeman:
I did find an issue while trying to update the fx/obs return values from sites. Trying to import the Forecast or Observation schemas throws an ImportError,  I think because of the import of siteschema creating a cycle. 

I could move the schema into their own module and each flask view module could import the ones they need to avoid the cycle?

### DIFF
--- a/sfa_api/forecasts.py
+++ b/sfa_api/forecasts.py
@@ -2,10 +2,10 @@ from flask import Blueprint
 from flask.views import MethodView
 
 
-from sfa_api import spec
-from sfa_api.schema import (ForecastValueSchema,
-                            ForecastSchema, ForecastResponseSchema,
-                            ForecastLinksSchema)
+from sfa_api import spec, ma
+from sfa_api.schema import SiteSchema, ForecastValueSchema, \
+                           ForecastPostSchema, ForecastSchema, \
+                           ForecastLinksSchema
 from sfa_api.demo import Forecast, TimeseriesValue
 
 
@@ -28,7 +28,7 @@ class AllForecastsView(MethodView):
           401:
             $ref: '#/components/responses/401-Unauthorized'
         """
-        forecasts = [Forecast() for i in range(3)]
+        forecasts = [ Forecast() for i in range(3)]
         return ForecastSchema(many=True).jsonify(forecasts)
 
     def post(self, *args):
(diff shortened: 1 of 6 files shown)


sfa_api/observations.py
```diff
@@ -46,15 +46,13 @@ class Meta:
     uuid = ma.UUID()
     _links = ma.Hyperlinks({
         'metadata': ma.AbsoluteURLFor('observations.metadata',
-                                      uuid='<uuid>'),
+                                      obs_id='<uuid>'),
```
alorenzo175: Maybe everything should be updated to have obs_id instead of uuid? (like in schemas above)

sfa_api/sites.py
```diff
@@ -206,12 +206,14 @@ def get(self, site_id, *args):
               application/json:
                 schema:
                   type: array
+                  items:
+                    $ref: '#/componens/schemas/ForecastMetadata'
```
alorenzo175: typo "componens" missing t

sfa_api/forecasts.py
```diff
@@ -0,0 +1,269 @@
...
+        summary: Add Forecast data
+        description: Add new timeseries values to Forecast entry.
+        tags:
+        - Forecasts
+        parameters:
+        - $ref: '#/components/parameters/forecast_id'
+        requestBody:
+          required: True
+          content:
+            application/json:
```
alorenzo175: can you add a csv example?

sfa_api/forecasts.py
```diff
@@ -0,0 +1,269 @@
...
+    site_id = ma.UUID(description="UUID of associated site",
+                      required=True)
+    name = ma.String(description="Human friendly name for forecast",
+                     required=True)
+
+
+@spec.define_schema('ForecastMetadata')
+class ForecastSchema(ForecastPostSchema):
+    site = ma.Nested(SiteSchema)
+    uuid = ma.UUID()
```
alorenzo175: like site_id, should we change these to forecast_id?

sfa_api/forecasts.py
lboeman: I reused the Observation example for this, the file format should be the same right?

alorenzo175: yup

sfa_api/forecasts.py
lboeman: I converted the instances of uuid to *_id it clarifies the code, although it can be confusing in the responses (e.g. forecasts come back with forecast_id and site_id attributes). But users should already know the response type and which id goes with the response object, so it shouldn't be an issue.

PR #24, opened 2018-11-09
add more intro material

### LINKED ISSUE #23: intro documentation
We should add more intro material that outlines how we expect most people to interact with the API: POSTing and GETing preexisting observations. An important item to clarify is one csv per observation (so one csv per data type) and therefore likely multiple POST/GET operations per site.

alorenzo175:
closes #23


alorenzo175:
anything else to add/change?

wholmgren:
looks good to me

PR #25, opened 2018-12-06
WIP: bring API in line with Data Model document

lboeman:
Starting updating the api to reflect the data model document. Marked work in progress until the data model is solidified.
Also added a /schema endpoint which returns a JSON Schema representation of the post schemas.


lboeman:
Closing because moving away from schema endpoint idea.

PR #28, opened 2019-01-31
Observation endpoint parsing

lboeman:
- Added parsing logic for post/get requests
- Add testing for observations endpoint
- Updated Schema to reflect latest Data Model version.
- Added utils/storage package as abstraction layer for SQL connection. This is more a conceptual stand-in than a suggestion for the concrete implementation of our sql/storage interface.

Haven't added a check for resource existence in the GET requests. Perhaps this check could be rolled into the auth check, where if a user does not have access to a resource or the resource does not exist, it fails and we return a 404?


sfa_api/observations.py
```diff
@@ -169,12 +197,64 @@ def post(self, obs_id, *args):
...
+        try:
+            observation_df['timestamp'] = pd.to_datetime(
+                observation_df['timestamp'],
+                utc=True)
+        except ValueError:
+            errors.append('Invalid item in "timestamp" field.')
+        except KeyError:
+            errors.append('Missing "timestamp" field.')
+
+        if 'questionable' not in observation_df.columns:
```
alorenzo175: questionable -> quality_flag here and everywhere

sfa_api/schema.py
```diff
@@ -1,35 +1,103 @@
...
+        description='Additional user specified parameters.')
+
+
 # Sites
+@spec.define_schema('ModelingParameters')
+class ModelingParameters(ma.Schema):
+    ac_power = ma.String(
+        title="AC Power",
+        description="Nameplate AC power rating.")
+    dc_power = ma.String(
```
alorenzo175: both powers should be a float

sfa_api/schema.py
```diff
@@ -38,7 +106,9 @@ class ObservationValueSchema(ma.Schema):
     class Meta:
         strict = True
         ordered = True
-    timestamp = ma.DateTime(description="ISO 8601 Datetime")
+    timestamp = ma.DateTime(
+        title="Timestamp",
+        description="ISO 8601 Datetime")
     value = ma.Float(description="Value of the measurement")
     questionable = ma.Boolean(description="Whether the value is questionable",
```
alorenzo175: quality_flag

alorenzo175: and an integer

sfa_api/schema.py
```diff
@@ -74,50 +160,85 @@ class Meta:
...
-    value = ma.Float(description="Value of the measurement")
-    questionable = ma.Boolean(description="Whether the value is questionable",
-                              default=False, missing=False)
+    timestamp = ma.DateTime(
+        title="Timestamp",
+        description="ISO 8601 Datetime")
+    value = ma.Float(
+        title="Value",
+        description="Value of the forecast variable.")
+    questionable = ma.Boolean(
```
alorenzo175: quality_flag int

sfa_api/schema.py
```diff
@@ -74,50 +160,85 @@ class Meta:
...
+    issue_time = ma.String(
+        title='Issue Time of Day',
+        required=True,
+        description=('The time of day that a forecast run is issued, '
+                     'e.g. 00:30. For forecast runs issued multiple '
+                     'times within one day (e.g. hourly), this specifies '
+                     'the first issue time of day. Additional issue times '
+                     'are uniquely determined by the first issue time and '
+                     'the run length & issue frequency attribute.'))
+    lead_time_to_start = ma.String(
```
alorenzo175: will need to verify this, duration, and issue freq later

alorenzo175: and issue_time. probably easiest to use pandas timedeltas for the former three, and a time function for issue_time

sfa_api/test/test_observation.py
```diff
@@ -0,0 +1,177 @@
...
+}
+VALID_CSV = "#I am a header comment, I am going to be ignored\ntimestamp,value,questionable\n2018-10-29T12:04:23Z,32.93,0\n2018-10-29T12:05:23Z,32.93,0\n2018-10-29T12:06:23Z,32.93,0\n2018-10-29T12:07:23Z,32.93,0\n" # NOQA
+WRONG_DATE_FORMAT_CSV = "timestamp,value,questionable\nksdfjgn,32.93,0"
+NON_NUMERICAL_VALUE_CSV = "timestamp,value,questionable\n2018-10-29T12:04:23Z,fgh,0" # NOQA
+NON_BINARY_FLAG_CSV = "timestamp,value,questionable\n2018-10-29T12:04:23Z,32.93,B" # NOQA
+
+
+# TODO: mock retrieval request to return a static observation for testing
+@pytest.mark.parametrize('uuid', [
+    ('7365da38-2ee5-46ed-bd48-c84c4cc5a6c8')
```
alorenzo175: since you reuse this lots of places, maybe make it a pytest fixture so we can updated in one place if needed

sfa_api/demo.py
```diff
@@ -1,26 +1,38 @@
...
+    well_known_text = None
+    modeling_parameters = {
+        "ac_power": "",
+        "axis_azimuth": 45.0,
+        "axis_tilt": 45.0,
+        "backtrack": True,
+        "dc_power": "",
+        "temperature_coefficient": "",
+        "ground_coverage_ratio": 0.5,
+        "surface_azimuth": 45.0,
```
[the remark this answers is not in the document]
wholmgren: pvlib convention is south = 180. let's stick with that. let's make sure we document that in all the right places (api, data model, dashboard tooltip)

lboeman:
I think this is ready, at least to update the documentation for stakeholder feedback. I believe I covered all of the comments and  have opened #29  and referenced some of the comments @alorenzo175  made so we can handle the forecast schema updates and validation in a different PR. Any objections to merging this?

sfa_api/demo.py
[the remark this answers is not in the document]
alorenzo175: Just make surface_azimuth = 0, then ok to merge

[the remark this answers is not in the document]
alorenzo175: meant 180 whoops

PR #39, opened 2019-02-28
development storage backend

lboeman:
Adds an in memory storage back end for demonstration or development purposes which won't persist between application instances.

I've added a file describing a storage 'interface' that may be better as an abstract base class that eventually gets extended with a sql connection for production. Perhaps it isn't necessary at all, but for the time being I think it is helpful to have something to program against.

Removed the static responses that were being returned by the current dev API hosted at [https://dev-api.solarforecastarbiter.org/](https://dev-api.solarforecastarbiter.org/), so those endpoints will no longer function. Should those remain up for the time being?

I also updated the requirements to use the pre-release version of marshmallow3 as recommended  here: [https://github.com/marshmallow-code/marshmallow/issues/1154](https://github.com/marshmallow-code/marshmallow/issues/1154). Marshmallow 3 does a better job of error reporting and handling nested schema.


lboeman:
Contains working endpoints for sites/observations/forecasts. This PR still needs updated unit tests, but could use some system testing.
Will run locally with:
`pip install -r requirements.txt
FLASK_APP=sfa_api` FLASK_ENV=development SFA_API_STATIC_DATA=true flask run -p <port>`

wholmgren:
is it expected that `curl localhost:5000/sites` and `curl http://localhost:5000/sites` return redirect notices? only `curl http://localhost:5000/sites/` worked. maybe I should have known better?

wholmgren:
the flask app crashed when I tried to delete a site that I created

lboeman:
> is it expected that `curl localhost:5000/sites` and `curl http://localhost:5000/sites` return redirect notices? only `curl http://localhost:5000/sites/` worked. maybe I should have known better?

Yes, endpoints that act on all resources of a type, e.g. sites, have a trailing slash. This is a typical albeit not required REST convention (see [this stack exchange post](https://softwareengineering.stackexchange.com/questions/186959/trailing-slash-in-restful-api)) .  I do agree that getting a 300 when you're trying to program against an API is a little confusing. Perhaps we need to explicitly state that people should expect this behavior in the API docs?

wholmgren:
>  I do agree that getting a 300 when you're trying to program against an API is a little confusing. Perhaps we need to explicitly state that people should expect this behavior in the API docs?

Yes, let's state it. I noticed that `requests` automatically adds the trailing slash.

wholmgren:
The `_links` attribute is super handy now that I'm trying to manually explore the api with curl. Nice.

wholmgren:
`POST observations/` restricts value type to "interval_mean" and "instantaneous" but the doc string suggests other options

wholmgren:
What's going on with this request to post observation values?

```
curl -D - http://localhost:5000/observations/b1dfe2cb-9c8e-43cd-afcf-c5a6feaf81e2/values -d @obsdata2.json -H 'content-type: application/json'

HTTP/1.0 400 BAD REQUEST
Supplied JSON does not contain "values" field.

#obsdata2.json = copy/paste of example data
[
  {
    "quality_flag": false,
    "timestamp": "2019-03-05T17:54:11Z",
    "value": 0
  }
]
```

Not clear to me why it doesn't work. Also should `Supplied JSON does not contain "values" field.` refer to `value` instead?

lboeman:
> What's going on with this request to post observation values?
> 
> ```
> curl -D - http://localhost:5000/observations/b1dfe2cb-9c8e-43cd-afcf-c5a6feaf81e2/values -d @obsdata2.json -H 'content-type: application/json'
> 
> HTTP/1.0 400 BAD REQUEST
> Supplied JSON does not contain "values" field.
> 
> #obsdata2.json = copy/paste of example data
> [
>   {
>     "quality_flag": false,
>     "timestamp": "2019-03-05T17:54:11Z",
>     "value": 0
>   }
> ]
> ```
> Not clear to me why it doesn't work. Also should `Supplied JSON does not contain "values" field.` refer to `value` instead?

There is a discrepancy between the Data Model and the API documentation, The API needs to be updated to reflect the correct json format for posting. The JSON payload should be be in the form
```
{
  'id': 'ignored-field',
  'values': [ ...data to post...]
}
``` 
Documented here: [https://solarforecastarbiter.org/datamodel/#json-1](https://solarforecastarbiter.org/datamodel/#json-1).  This will allow us to do things like clone data easily since the format will match the response for GET requests.

wholmgren:
Maybe a bigger question than this PR, but does deleting a site delete all associated observations and forecasts (given sufficient permissions)?

lboeman:
> Maybe a bigger question than this PR, but does deleting a site delete all associated observations and forecasts (given sufficient permissions)?

I think we'd have to if we're not allowing metadata edits. It seems like reassigning sites on observations/forecasts would invalidate data as well.

wholmgren:
(csv seems to be a separate issue so moving to its own comment...)

Also does not work with csv files:

```
curl -D - http://localhost:5000/observations/b1dfe2cb-9c8e-43cd-afcf-c5a6feaf81e2/values -d @obsdata.csv -H 'content-type: text/csv'

HTTP/1.0 400 BAD REQUEST

{
  "errors": [
    "Missing \"quality_flag\" field."
  ]
}

# obsdata.csv
timestamp,value,quality_flag
2019-01-01T12:00:00+00:00,27.95791511063986,10
2019-01-01T12:01:00+00:00,21.82482342812345,7
2019-01-01T12:02:00+00:00,54.884324244538504,8
2019-01-01T12:03:00+00:00,24.71352737530024,6
2019-01-01T12:04:00+00:00,46.37681546831439,1
```

Strange that the error message is formatted differently for csv (plain text vs. json).

I also get `400... Malformed CSV` responses if I add a `# header` line.

lboeman:
> Strange that the error message is formatted differently for csv (plain text vs. json).
I added #40 to address this. I'm leaning towards always returning errors in JSON format because it's human readable and easy to parse.

sfa_api/demo/sites.py
```diff
@@ -0,0 +1,50 @@
...
+        "elevation": 786.0,
+        "extra_parameters": {
+            '{"network": "NREL MIDC"}',
+        },
+        "latitude": 32.22969,
+        "longitude": -110.95534,
+        "modeling_parameters": {},
+        "name": "Weather Station 1",
+        "provider": "Reference",
+        "timezone": "America/Pheonix",
```
wholmgren: `Pheonix -> Phoenix`

sfa_api/demo/sites.py
```diff
@@ -0,0 +1,50 @@
+static_sites = {
+    '123e4567-e89b-12d3-a456-426655440001': {
+        "elevation": 595.0,
+        "extra_parameters": {
+            '{"net_work_api_abbreviation": "AS","network": "University of Oregon SRML","network_api_id": "94040"}', # NOQA
```
wholmgren: net_work -> network

wholmgren:
forecasts/values should not contain `quality_flag`
```
curl -D - http://localhost:5000/forecasts/f8dd49fa-23e2-48a0-862b-ba0af6dec276/values
[
  {
    "quality_flag": 0,
    "timestamp": "2019-01-01T12:00:00+00:00",
    "value": 66.23864379696639
  },...
```

wholmgren:
format for all time-related parameters should be documented like for issue time of day

wholmgren:
forecast creation does not return new UUID:

```
curl -D - http://localhost:5000/forecasts/ -d @fxmeta.json -H 'content-type: application/json'
HTTP/1.0 201 CREATED

Forecast created.
```

wholmgren:
Is the test storage supposed to be able to save forecast time series data? It does not.

lboeman:
> forecasts/values should not contain `quality_flag`
> 
> ```
> curl -D - http://localhost:5000/forecasts/f8dd49fa-23e2-48a0-862b-ba0af6dec276/values
> [
>   {
>     "quality_flag": 0,
>     "timestamp": "2019-01-01T12:00:00+00:00",
>     "value": 66.23864379696639
>   },...
> ```
> I see this was part of the 0.1.0 version spec, so I guess I just missed earlier. However, it's not part of the dashboard interaction example: https://solarforecastarbiter.org/dashboarddoc/#upload-data

This is an oversight on my part, I'm not supplying quality_flags but the copy pasta'd Forecast Schema includes a quality_flag field with default 0, so it adds it when it serializes the python objects to JSON.

lboeman:
> curl -D - http://localhost:5000/observations/b1dfe2cb-9c8e-43cd-afcf-c5a6feaf81e2/values -d @obsdata.csv -H 'content-type: text/csv'

It looks like the csv received by the API from this curl command doesn't have any newlines in it, so it reads the header line as 'timestamp,value,quality_flag2019-01-01T12:00:00+00:00' resulting in the missing quality flag field error. Using the option --data-binary will send newlines as well. 
`curl -D - http://localhost:5000/observations/b1dfe2cb-9c8e-43cd-afcf-c5a6feaf81e2/values --data-binary @obsdata.csv -H 'content-type: text/csv'`.

lboeman:
@wholmgren  I believe I have covered all of your points here and this is in decent shape to merge. Any objections?

wholmgren:
Summary of concerns:

- [ ] Getting a 300 when you're trying to program against an API is a little confusing. Perhaps we need to explicitly state that people should expect this behavior in the API docs?
- [ ] the flask app crashed when I tried to delete a site that I created
- [ ] POST observations/ restricts value type to "interval_mean" and "instantaneous" but the doc string suggests other options
- [ ] There is a discrepancy between the Data Model and the API documentation, The API needs to be updated to reflect the correct json format for posting. 
- [ ] Maybe a bigger question than this PR, but does deleting a site delete all associated observations and forecasts (given sufficient permissions)?
- [ ] forecasts/values should not contain quality_flag
- [ ] format for all time-related parameters should be documented like for issue time of day
- [ ] forecast creation does not return new UUID

Can you check off the ones you've done so that I don't have to go confirm it myself?

sfa_api/observations.py
```diff
@@ -273,8 +283,12 @@ def post(self, obs_id, *args):
 
         if errors:
             return jsonify({'errors': errors}), 400
-
+        index = observation_df['timestamp'].copy()
+        observation_df = observation_df.set_index(index)
```
alorenzo175: same set_index as above

sfa_api/forecasts.py
```diff
@@ -261,8 +274,11 @@ def post(self, forecast_id, *args):
 
         if errors:
             return jsonify({'errors': errors}), 400
-
+        forecast_df = forecast_df.set_index(forecast_df['timestamp'].copy())
```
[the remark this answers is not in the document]
alorenzo175: although should drop be True so the 'timestamp' column is removed. Seems like forecasts in the storage layer above are supposed to only have value, quality_flag columns

sfa_api/sites.py
```diff
@@ -26,18 +27,22 @@ def get(self, *args):
                   items:
                     $ref: '#/components/schemas/SiteMetadata'
           401:
-            $ref: '#/components/responses/401-Unauthorized'
+            dref: '#/components/responses/401-Unauthorized'
```
alorenzo175: supposed to have changed?

lboeman: Nope! I'll fix that, thanks!

lboeman:
> Getting a 300 when you're trying to program against an API is a little confusing. Perhaps we need to explicitly state that people should expect this behavior in the API docs?

I've added ' Note that POST requests to this endpoint without a trailing slash will result in a redirect response.'  to the metadata post endpoints, do you think that is sufficient?

> Maybe a bigger question than this PR, but does deleting a site delete all associated observations and forecasts (given sufficient permissions)?

I've set the test storage to do so for the time being. It seems reasonable to do per the data model, but it seems jarring for users if someone can delete a Site and destroys their data.

I'll make another commit to fix the documentation issues.

wholmgren:
> ...to the metadata post endpoints, do you think that is sufficient?

Yes

> it seems jarring for users if someone can delete a Site and destroys their data.

Dashboard deletion procedure should probably have a "are you sure you want to do this?" question. Otherwise it's fine for now.

lboeman:
> POST observations/ restricts value type to "interval_mean" and "instantaneous" but the doc string suggests other options

Removed the docstring suggesting others, options are listed in an array-like notation on the documentation now, so if we add options to the `VALUE_TYPES` variable in schema.py they'll be included in the documentation.

> There is a discrepancy between the Data Model and the API documentation, The API needs to be updated to reflect the correct json format for posting.

Created wrapper schemas Observation/ForecastValuesSchema for responses to follow the format in the data model and dashboard docs.

> * format for all time-related parameters should be documented like for issue time of day

Wrote a validator, TimeFormat in sfa_api.utils.validators.py that verifies time values are all in the HH:MM format. 

I think this is _finally_ in a state to merge.

sfa_api/utils/validators.py
```diff
@@ -0,0 +1,18 @@
...
+class TimeFormat(Validator):
+    """Validates Time is in correct format"""
+    def __init__(self, time_format):
+        self.time_format = time_format
+
+    def __call__(self, value):
+        try:
+            time.strptime(value, self.time_format)
+        except ValueError:
+            raise ValidationError('Time not in HH:MM format.')
```
wholmgren: maybe `f'Time not in {self.time_format} format.'`

wholmgren:
@lboeman I just enabled the codecov integration for this repository. not sure if it will require a new push to work.

wholmgren:
I looked at the coverage report and I think its fine.

Is this a valid upload?

```
timestamp,value,quality_flag
20190301T120000Z,5,0
20190301T120000Z,,1
20190301T120000Z,5,0
20190301T120000Z,NaN,1
20190301T120000Z,NULL,1
```

wholmgren:
I suggest that "Invalid item in \"timestamp\" field." becomes "Invalid item in \"timestamp\" field. Ensure that timestamps are ISO8601 compliant"

wholmgren:
value: "Invalid item in \"value\" field. Ensure that all values are integers, floats, empty, NaN, or NULL"

quality flag: "Invalid item in \"quality_flag\" field. Ensure that all quality_flags are 0 or 1"

wholmgren:
the accept: text/csv response is garbage
`curl -D - http://localhost:5000/observations/b1dfe2cb-9c8e-43cd-afcf-c5a6feaf81e2/values -H 'accept: text/csv'`

wholmgren:
Starting with fresh API instance, the following leads to a crash presumably due to faulty handling of missing data:

```
$ curl -D - http://localhost:5000/observations/b1dfe2cb-9c8e-43cd-afcf-c5a6feaf81e2/values
# valid json response

# where obsdata is as posted above with missing data
$ curl -D - http://localhost:5000/observations/b1dfe2cb-9c8e-43cd-afcf-c5a6feaf81e2/values --data-binary @obsdata.csv -H 'content-type: text/csv'

$ curl -D - http://localhost:5000/observations/b1dfe2cb-9c8e-43cd-afcf-c5a6feaf81e2/values
marshmallow.exceptions.ValidationError: {'values': {21: {'value': ['Special numeric values (nan or infinity) are not permitted.']}, 23: {'value': ['Special numeric values (nan or infinity) are not permitted.']}, 24: {'value': ['Special numeric values (nan or infinity) are not permitted.']}}}

-->
```

wholmgren:
Fixed the bad times in the test data:

```
timestamp,value,quality_flag
2019-03-01T12:00:00Z,5,0
2019-03-01T12:00:01Z,,1
2019-03-01T12:00:02Z,5,0
2019-03-01T12:00:03Z,NaN,1
2019-03-01T12:00:04Z,NULL,1
```

same upload/download roundtrip error as above

lboeman:
Updated the errors to those suggested, and corrected the problem I introduced that was mangling CSV responses, and correctly handle missing CSV values. All of the Curl requests are now returning the expected results. 
This will parse JSON null values into NaNs on the backend. I opened an issue about possibly expanding this later to accept other NaN values for JSON because many JSON serialization libraries don't export NaN/None as the Javascript null values, so it seems weird to expect it in a request but not a response.

If there are any other issues I should catch them while working on the dashboard and can work up a new PR. I think this is stable enough to merge. I'll keep a list of test cases to add to integration testing as I go. and collect them in #48.

wholmgren:
nice job, this was a big step forward.

PR #50, opened 2019-03-19
database initalization

alorenzo175:
This PR does a few things:
1. Starts defining database schema via that can be used with [migrate](https://github.com/golang-migrate/migrate)
2. defines the RBAC tables and triggers to appropriately update permissions
3. adds a docker-compose.yml file so that all of this can be tested
4. adds tests with pytest

It doesn't automate starting test mysql database, and some tests/schemas might be missing or incomplete


alorenzo175:
I think this is ready for a quick review, and I'll add more in smaller PRs. Not sure who should look at it. @wholmgren  @lboeman

lboeman:
I can take a look at this.

alorenzo175:
Yes, if someone wants to revoke a permission, like readX, from a user and that user has multiple roles with the readX permission, all those roles would have to be removed from the user so the user can not readX.

wholmgren:
does this close #27?

alorenzo175:
partly closes #27, I would consider it closed once all the functions we need are there and there is some test data in it. That should all be ready in the next day or two

Issue #38, opened 2019-02-25
Working Test API from static content in demo module.


lboeman:
- [x] create demo objects to act as static data store.
- [x] Make storage module to load and interact with static data store.
- [ ] Update dashboard to call API for data.

issues discovered while working on dashboard:
- [x] Observations lack `interval_length` field 
- [x] `site_id` stored as UUID when parsed by marshmallow. cast to string for storage.
- [x] parse and store time duration variables in minutes
- [x] don't allow duplicate timestamps in uploads.

lboeman:
@alorenzo175  @wholmgren  I've found some more time validation that will need to be added to the API. I'm not sure what reasonable bounds should be for the forecast time duration variables. We're currently parsing time duration variables explicitly for %H:%M format.

- [ ] issue_time of day: are all values between 00:00 and 23:59 valid?
- [ ] interval_length and run_length: should not be 00:00 Are there other bounds?
- [ ] lead_time_to_start: This will probably need a different validator, as the current one cannot validate values  >23:59. Do we need lead time to be this granular or would a set of values be sufficient? 

These conditions should be added to tests as well as the integration tests #48

wholmgren:
issue time of day: yes for now.

others: one option could be that the API requires all timedeltas in minutes and the dashboard does the conversion. I think we might have discussed that together at some point.

lboeman:
> issue time of day: yes for now.
> 
> others: one option could be that the API requires all timedeltas in minutes and the dashboard does the conversion. I think we might have discussed that together at some point.

I see I misunderstood a comment in my last PR, I mistakenly changed the expected format of timedeltas to HH:MM. I thought we'd just be storing timedeltas in minutes, but I think that requiring users to post in minutes is a better solution. I'll update the API to accept integer timedeltas in minutes.

alorenzo175:
Seems like enough of this has been implemented?

lboeman:
Agreed.

PR #51, opened 2019-03-19
Post validation

lboeman:
Supports #38 

1. Adds missing interval_length field to observation schema
2. Correct development storage backend storing keys as UUID objects when nested in other schema.
3. Unifies error formatting. There are some D.R.Y. issues at the moment. I think this should be handled by creating some custom exceptions and flask error handling. This needs a little more thought to continue to provide feedback for _all_ of the errors a user should fix instead of the first one encountered.
4. Change Time delta fields to integers.

### DIFF
--- a/sfa_api/schema.py
+++ b/sfa_api/schema.py
@@ -185,10 +185,15 @@ class ObservationPostSchema(ma.Schema):
                      'instant for instantaneous data'),
         validate=validate.OneOf(['beginning', 'ending', 'instant']),
         required=True)
+    interval_length = ma.String(
+        title='Interval length',
+        description=('The length of time that each data point represents  in'
+                     'HH:MM format, e.g. 00:05 for 5 minutes.'),
+        validate=TimeFormat('%H:%M'),
+        required=True)
     value_type = ma.String(
         title='Value Type',
-        validate=validate.OneOf(VALUE_TYPES)
-    )
+        validate=validate.OneOf(VALUE_TYPES))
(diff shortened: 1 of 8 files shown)


sfa_api/test/test_forecast.py
```diff
@@ -8,9 +8,9 @@
     "variable": "ac_power",
     "interval_label": "beginning",
     "issue_time_of_day": "12:00",
-    "lead_time_to_start": "01:00",
-    "interval_length": "00:01",
-    "run_length": "01:00",
+    "lead_time_to_start": 60,
+    "interval_length": 1,
+    "run_length": 86400,
```
alorenzo175: I think this should probably be different like 1440 (1 day in minutes instead of 60 days). same for demo/forecasts.py

sfa_api/schema.py
```diff
@@ -10,7 +10,8 @@
              'poa', 'ac_power', 'dc_power', 'pdf_probability',
              'cdf_value']
 
-VALUE_TYPES = ['interval_mean', 'instantaneous']
+VALUE_TYPES = ['interval_mean', 'interval_max', 'interval_min',
+               'interval_media', 'percentile', 'instantaneous']
```
alorenzo175: oop just saw the `interval_media` typo

lboeman: Woops. Thanks. This should probably get broken up as well, because observations and forecasts have different allowable value types :/

PR #53, opened 2019-03-21
update database schemas for sites, forecasts, observations

alorenzo175:
after #52 is merged


alorenzo175:
I've set many metadata fields to DECIMAL to avoid float conversion changing values. Are the precision of the DECIMAL fields enough? DECIMAL(10, 6) means there will be 6 digits after the decimal and 4 before. Did I miss any parameters?  @wholmgren

datastore/migrations/0003_add_other_tables.up.sql
```diff
@@ -3,7 +3,27 @@ CREATE TABLE arbiter_data.sites (
     id BINARY(16) NOT NULL DEFAULT (UUID_TO_BIN(UUID(), 1)),
     organization_id BINARY(16) NOT NULL,
     name VARCHAR(64) NOT NULL,
-    -- add everything else later...
+    latitude DECIMAL(8, 6) NOT NULL,
+    longitude DECIMAL(9, 6) NOT NULL,
```
wholmgren: negative ok in this spec?

datastore/migrations/0003_add_other_tables.up.sql
```diff
@@ -3,7 +3,27 @@ CREATE TABLE arbiter_data.sites (
...
+    elevation DECIMAL(7, 2) NOT NULL,
+    timezone VARCHAR(32) NOT NULL,
+    extra_parameters VARCHAR(255) NOT NULL,
+    ac_capacity DECIMAL(10, 5) NOT NULL,
+    dc_capacity DECIMAL(10, 5) NOT NULL,
```
wholmgren: 0.00001 MW to 99999.99999 MW seems ok. add a digit and no one would ever complain.

datastore/migrations/0003_add_other_tables.up.sql
```diff
@@ -18,8 +38,14 @@ CREATE TABLE arbiter_data.observations (
...
     site_id BINARY(16) NOT NULL,
     name VARCHAR(64) NOT NULL,
-
-    -- add everything else later...
+    variable VARCHAR(32) NOT NULL,
+    interval_label ENUM('beginning', 'ending', 'instant') NOT NULL,
+    interval_length SMALLINT UNSIGNED NOT NULL,
+    value_type VARCHAR(32) NOT NULL,
+    uncertainty FLOAT NOT NULL,
+    extra_parameters VARCHAR(255) NOT NULL,
```
wholmgren: only 255 characters? I can imagine a verbose metadata spec bumping into that

datastore/migrations/0003_add_other_tables.up.sql
alorenzo175: yes, all decimal fields are +- the range

datastore/migrations/0003_add_other_tables.up.sql
wholmgren: more likely for a site than an obs, commented in the wrong place

datastore/migrations/0003_add_other_tables.up.sql
```diff
@@ -18,8 +38,14 @@ CREATE TABLE arbiter_data.observations (
     organization_id BINARY(16) NOT NULL,
     site_id BINARY(16) NOT NULL,
     name VARCHAR(64) NOT NULL,
-
-    -- add everything else later...
+    variable VARCHAR(32) NOT NULL,
```
wholmgren: did you consider `enum` here? I think enum would prove to be overly restrictive, but it's consistent with the current data model

datastore/migrations/0003_add_other_tables.up.sql
```diff
@@ -3,7 +3,27 @@ CREATE TABLE arbiter_data.sites (
...
+    temperature_coefficient DECIMAL(7, 5) NOT NULL,
+    tracking_type ENUM('fixed', 'single_axis') NOT NULL,
+    surface_tilt DECIMAL(4, 2),
+    surface_azimuth DECIMAL(5, 2),
+    axis_tilt DECIMAL(4, 2),
+    axis_azimuth DECIMAL(5, 2),
+    ground_coverage_ratio DECIMAL(8, 4),
+    backtrack BOOLEAN,
+    max_rotation_angle DECIMAL(5, 2),
+    irradiance_loss_factor DECIMAL(5, 2) NOT NULL,
```
wholmgren: not supporting this

datastore/tests/conftest.py
```diff
@@ -95,16 +96,33 @@ def fcn(action, object_type, applies_to_all, org=None):
...
-            'INSERT INTO sites (id, organization_id, name) VALUES '
-            '(%s, %s, %s)', list(out.values()))
+                          name=f'site{str(uuid1())[:10]}',
+                          latitude=0, longitude=0, elevation=0,
+                          timezone='America/Denver', extra_parameters='',
+                          ac_capacity=0, dc_capacity=0,
+                          temperature_coefficient=0,
+                          tracking_type='fixed',
+                          surface_tilt=0, surface_azimuth=0,
+                          axis_tilt=0, axis_azimuth=0,
```
wholmgren: if `tracking_type='fixed'` should the single axis parameters be `NULL`? Is that allowed in the schema?

datastore/migrations/0003_add_other_tables.up.sql
alorenzo175: considered but didn't want to have to update the schema if we add things later

datastore/tests/conftest.py
alorenzo175: yes, and it is allowed

alorenzo175: overall, want to try and avoid allowing a column to be null because it requires extra space

wholmgren: space as in memory? Aren't we talking about a few bytes per site? Seems worthwhile to promote internal data model consistency.

datastore/migrations/0003_add_other_tables.up.sql
```diff
@@ -3,7 +3,26 @@ CREATE TABLE arbiter_data.sites (
     id BINARY(16) NOT NULL DEFAULT (UUID_TO_BIN(UUID(), 1)),
     organization_id BINARY(16) NOT NULL,
     name VARCHAR(64) NOT NULL,
-    -- add everything else later...
+    latitude DECIMAL(8, 6) NOT NULL,
+    longitude DECIMAL(9, 6) NOT NULL,
+    elevation DECIMAL(7, 2) NOT NULL,
+    timezone VARCHAR(32) NOT NULL,
+    extra_parameters TEXT NOT NULL,  -- unclear if this will have a positive or negative effect on performance
```
wholmgren: I thought a limit was reasonable, just something bigger than 255 characters. We should put a limit in the API.

datastore/tests/conftest.py
alorenzo175: I meant avoid null in general, but these parameters are allowed to be null to be consistent. Space as in memory and disk.

alorenzo175: and define then (now as None) in conftest for easier comparison to result from DB later

datastore/migrations/0003_add_other_tables.up.sql
alorenzo175: I agree about an API limit, but what limit is reasonable? 
```
This webpage documents the public Solar Forecast Arbiter API. This RESTful API is primarily meant to be accessed by the Solar Forecast Arbiter dashboard. The API relies primarily on JSON data structures in requests and responses, with the notable exception of the forecast and observation GET/POST data endpoints which also support CSV files
```
is already 341 characters

alorenzo175: I think it's ok to let the api set the limit and store whatever it gives to the db

PR #54, opened 2019-03-21
add db tests to pipeline

alorenzo175:


alorenzo175:
I give up, will just have to do database ci on openshift me thinks.

PR #62, opened 2019-03-25
add DB procedures to insert data

alorenzo175:
currently can only insert one row at a time. A bulk insert might be necessary, although tougher to secure, based on how slow this implementation is.


alorenzo175:
I think all ids are stored as BINARY(16) including foreign keys. These interface prcedures all accept string IDs and convert with UUID_TO_BIN(?, 1) (which rearranges the time part of the UUID for more efficient indexing). I want to stick with the API dealing only with string IDs and let MySQL convert if necessary.

PR #68, opened 2019-03-26
add demo data to DB

alorenzo175:
and add a small test implementation of ``storage_interface`` talking to the db. Also updated the demo objects


alorenzo175:
I don't want this to get too big, so I'll continue in a new PR. One thing to review here is if we want to always return the full modeling parameters dict with None  values. I think it makes sense to always have the same keys in the json response

Issue #56, opened 2019-03-22
Include metadata links in value responses


lboeman:
Add a metadata link to csv header comments (as well as uuid) and _links attribute to json.

Issue #40, opened 2019-03-05
Decide on Error Handling


lboeman:
We'll want to define how errors are reported in responses so that responses are easily consumable (e.g. json returned to the dashboard for display). Probably worth implementing some custom exceptions and error handlers for Flask as shown here: [http://flask.pocoo.org/docs/1.0/patterns/apierrors/](http://flask.pocoo.org/docs/1.0/patterns/apierrors/)

lboeman:
closed by #73

PR #69, opened 2019-03-27
CDF implementation

lboeman:
Adds CDF Forecast endpoints to the  API and the mock database.


lboeman:
The code here assumes the storage class is responsible for creating the individual cdf_forecasts in the group, because to store them we will need the forecast_id of their cdf_forecast_group for the 'parent' field, and the cdf_forecast group will need to convert constant_values from the user-provided values to references to the cdf_forecasts.

sfa_api/demo/__init__.py
```diff
@@ -336,3 +344,255 @@ def list_sites():
...
+    dict
+        The CDF Forecast's metadata or None if the Forecast
+        does not exist.
+    """
+    if forecast_id not in cdf_forecast_groups:
+        return None
+    else:
+        forecast_group = cdf_forecast_groups[forecast_id]
+        with_children = forecast_group.copy()
+        children = [cdf_forecasts[fx_id]
```
alorenzo175: can we basically do this in demo/cdf_forecasts.py? for easier testing w/ database

sfa_api/demo/cdf_forecasts.py
```diff
@@ -0,0 +1,103 @@
...
+        "site_id": "123e4567-e89b-12d3-a456-426655440001",
+        "variable": "ghi",
+        "issue_time_of_day": "06:00",
+        "interval_length": 5,
+        "run_length": 1440,
+        "interval_label": "beginning",
+        "lead_time_to_start": "60",
+        "interval_value_type": "interval_mean",
+        "axis": "y",
+        "constant_values": ['633f9396-50bb-11e9-8647-d663bd873d93',
```
alorenzo175: so could define static_cdf_forecasts first, then pull the values from that dict into this list

sfa_api/forecasts.py
```diff
@@ -380,7 +388,21 @@ def post(self, *args):
...
+        data = request.get_json()
+        try:
+            cdf_forecast_group = CDFForecastGroupPostSchema().load(data)
+        except ValidationError as err:
+            return jsonify({"errors": err.messages}), 400
+        else:
+            storage = get_storage()
+            forecast_id = storage.store_cdf_forecast_group(cdf_forecast_group)
+            if forecast_id is None:
+                return jsonify({'errors': 'Site does not exist'}), 400
```
alorenzo175: we'll need to change these later to 404 or 401. Should have the same message if the user can't perform the action or the site does not exist

sfa_api/forecasts.py
```diff
@@ -500,7 +567,60 @@ def post(self, forecast_id):
           404:
             $ref: '#/components/responses/404-NotFound'
         """
-        return jsonify({'errors': {'error': ['Not Implemented']}})
+        if request.content_type == 'application/json':
```
alorenzo175: most of this should be shared with other forecast post, right? so make a utility function

sfa_api/sites.py
```diff
@@ -192,6 +193,39 @@ def get(self, site_id, *args):
         return jsonify(ForecastSchema(many=True).dump(forecasts))
 
 
+class SiteCDFForecastGroups(MethodView):
+    def get(self, site_id, *args):
+        """
+        ---
+        summary: Get site probabilistic forecast groups
+        description: >
+          Get metadata for all forecasts associated with site that
```
alorenzo175: all **CDF** forecasts

sfa_api/sites.py
```diff
@@ -192,6 +193,39 @@ def get(self, site_id, *args):
...
+          user has access to.
+        tags:
+        - Sites
+        parameters:
+        - $ref: '#/components/parameters/site_id'
+        responses:
+          200:
+            description: Successfully retrieved site forecasts
```
alorenzo175: cdf

sfa_api/forecasts.py
lboeman: I opened #56 for this, I think in this case it would always be a 404 because authentication would get checked before taking any action in the case of a post. I would like to make a PR with some custom exceptions to handle all the formatting as in #40 and I think we're far enough along now that I have a good enough idea of what we'll need to do this.

lboeman:
> So all the duplicate mimetype stuff will be moved later? Looks good. Just needs tests to merge (maybe demo/ should be added to .coveragerc?)

Yes, My plan is roll that into a function which parses a json or csv into a dataframe and we could use it for the forecast, cdf forecast and observation endpoints, and I'd like to implement an error handler for that so it could be a significant amount of code that may be better in another PR?
I'll add tests for the cdf forecasts so we can get this merged.

lboeman:
Alright, tests have been added and with sfa_api/demo added to omit in .coveragerd coverage is pretty good. My plan of action is to work on refactoring the post parsing, implement better error handling so that jsonify({'error':{..}}) isn't everywhere, and update tests to reflect these changes. I think this PR is ready to merge.

wholmgren:
looks good to me

PR #71, opened 2019-03-28
implement storage_interface for DB data

alorenzo175:
except for CDF stuff. Also switch from storing issue_time_of_day as TIME to string


alorenzo175:
Don't know if it's worth it to test the database separately instead of all together

alorenzo175:
I can wait to merge after rebase on #69 if that will be done soon

PR #73, opened 2019-03-29
Errors & some refactoring

lboeman:
Refactored shared request parsing into utility functions and introduce some exception handling so that we can raise exceptions from anywhere in the package in order to send Error responses.

### DIFF
--- a/sfa_api/utils/errors.py
+++ b/sfa_api/utils/errors.py
@@ -1,15 +1,30 @@
 class BaseAPIException(Exception):
-    """Will accept a dictionary of {error_name: description} or will
-    build a dictionary from kwargs. Both types of arguments can be
-    provided and will be joined. Wraps any non-list descriptions
-    in a list to mimic  the format of marshmallow errors.
+    """Base exception to be thrown from within API code to trigger
+    an immediate HTTP response.
 
     Parameters
     ----------
-    errors: Dict
-        A dictionary of errors to reply with. Optionally, key word
-        arguments can be provided instead of a dictionary.
-    kwargs:
+    status_code: int
+        The HTTP status code to return with the errors.
+
+    errors: dict
+        A dictionary where keys are names for an error and values are
+        a description. 
+        e.g {'Longitude': 'Must be between -180 and 180.'}
+
+    Notes
+    -----
+    Errors can be provided as a dictionary as described above, as
+    keyword arguments or both. Any non-list descriptors will be
+    wrapped in a list such that the dictionary:
+
+        {'error0': 'error message',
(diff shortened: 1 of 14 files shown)


lboeman:
There was a marshmallow update, and do to the fact that we're using pre-release 3.0.0, it broke my tests. However, the breaking change is actually a very nice feature. For enumerated variables, API errors return "Must be one of <list of choices>" rather than "Not a valid choice". So I'll fix these failing tests. 
@alorenzo175, @wholmgren  Do you think it would be better to set the specific prerelease version during development, or leave it and fix errors as they are introduced?

alorenzo175:
I think we should fix the tests and pin the version for now and make an issue to update the version once 3.0.0 is released

sfa_api/utils/errors.py
```diff
@@ -1,3 +1,36 @@
...
+        A dictionary of errors to reply with. Optionally, key word
+        arguments can be provided instead of a dictionary.
+    kwargs:
+    """
+    def __init__(self, status_code, errors={}, **kwargs):
+        Exception.__init__(self)
+        errors.update(kwargs)
+        for key, error in errors.items():
+            if not isinstance(error, list):
+                errors[key] = [error]
```
alorenzo175: what if error was a tuple?  Should probably test this module

lboeman: I don't think that would be a problem. What is your concern about supplying a tuple? Is providing errors as tuples common? 
I'll put some tests together for this.

alorenzo175: I figure you want errors to be list-like, so if error was a tuple (error0, erro1) instead of list after this block it would be [(error0, error1)]. 

I don't know if this would happen, just have to be careful when doing isinstance(list)

lboeman: Gotcha, that is definitely what's happening here. I was simply trying to ensure the top level data type of each value was a list, since the errors should really only be printed either as json in the browser or in the dashboard. This could use more documentation to explain that this is intended behavior though.

sfa_api/utils/tests/test_errors.py
```diff
@@ -0,0 +1,45 @@
...
+def test_badapirequest_dict():
+    exc = BadAPIRequest(error_dict)
+    assert exc.status_code == 400
+    assert exc.errors['error'] == ['message']
+    assert exc.errors['error2'] == ['message']
+    assert exc.errors['error3'] == [('went', 'wrong')]
+
+
+def test_basapirequest_kwargs():
+    exc = BadAPIRequest(error_dict)
```
alorenzo175: **error_dict?

PR #76, opened 2019-04-02
require authentication on all blueprint endpoints

alorenzo175:
and add tests for auth.py


alorenzo175:
Tests will fail until after 1:45 pm when the test token expires

sfa_api/utils/tests/test_auth.py
```diff
@@ -0,0 +1,80 @@
...
+        data=('{"grant_type": "password", '
+              '"username": "testing@solarforecastarbiter.org",'
+              '"password": "Thepassword123!", '
+              '"audience": "https://api.solarforecastarbiter.org", '
+              '"client_id": "c16EJo48lbTCQEhqSztGGlmxxxmZ4zX7"}'))
+    if token_req.status_code != 200:
+        pytest.skip('Cannot retrieve valid Auth0 token')
+    else:
+        token = token_req.json()['access_token']
+        return token
```
lboeman: Could this be moved into conftest with valid_auth for use in 'integration' tests? Maybe it's not totally useful there but it seems like if we're testing the whole stack this should be part of it, even if its redundant.

alorenzo175: sure, I don't see a problem with that

PR #84, opened 2019-04-04
add more test data to db

alorenzo175:
with a reference org that the current testing user can only read from


wholmgren:
did this break the dev site despite the passing tests?

alorenzo175:
nope, are you getting a 500 error? probably need to go to /logout and then log back in. Right now we don't have refresh tokens and the errors this causes aren't caught

wholmgren:
Ah that must be it.

PR #87, opened 2019-04-09
remove cdf_value, update air temperature variable name

### LINKED ISSUE #86: add poa_global to allowed variables

lboeman:
closes #86 
Removes cdf_value from allowed variables in schema.py
The dashboard is throwing a keyError trying to convert 'Air Temperature' so I updated it in the migrations sql to the variable name in the schema. @alorenzo175, would mysql need to be manually rebuilt in openshift for this change to take effect?


alorenzo175:
yes, once this is merged, you can just restart the mysql pod and it should update

PR #92, opened 2019-04-12
more schema validation

### LINKED ISSUE #80: validate modelingparameters schema
Should check what kind of tracking type is set and validate that the required fields for that tracking type are present

### LINKED ISSUE #90: String validation
We should validate things like site name better to avoid storing dangerous SQL injection or xss strings in the DB. Only allow letters digits and a few punctuation marks

alorenzo175:
closes #80 to validate modeling parameters depending on tracking type. Also closes #90 with a restricted set of allowed characters currently unicode word (a-z0-9_), comma, apostrophe, and space

### DIFF
--- a/sfa_api/conftest.py
+++ b/sfa_api/conftest.py
@@ -32,9 +32,9 @@ VALID_SITE_JSON = {
     "modeling_parameters": {
         "ac_capacity": 0.015,
         "dc_capacity": 0.015,
-        "backtrack": True,
+        "ac_loss_factor": 0,
+        "dc_loss_factor": 0,
         "temperature_coefficient": -.002,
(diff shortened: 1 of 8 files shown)


alorenzo175:
Ah, yeah I thought about that too but forgot to mention it. I think we should basically allow anything but be very careful when doing anything like rendering it in a template

wholmgren:
Could you add colon and dash and then apply same set to extra parameters? I don’t think anything else would be needed.

alorenzo175:
what if they want to do json? {}[]
or xml? <>

wholmgren:
Right... I could see people wanting to use () in names.

alorenzo175:
#93

PR #103, opened 2019-04-30
database user/role managment procedures

alorenzo175:


datastore/migrations/0012_role_management.up.sql
```diff
@@ -0,0 +1,398 @@
...
+    SET userid = UUID_TO_BIN(user_id, 1);
+    SET allowed = can_user_perform_action(auth0id, userid, 'update');
+    IF allowed THEN
+    INSERT INTO arbiter_data.user_role_mapping (user_id, role_id) VALUES (
+        userid, roleid);
+    ELSE
+        SIGNAL SQLSTATE '42000' SET MESSAGE_TEXT = 'Access denied to user on "add permission to role"',
+        MYSQL_ERRNO = 1142;
+    END IF;
+END;
```
lboeman: Here and in the `remove_role_from_user` procedure below, the permission checked is whether or not the originator of the request can update the user. I think this works inside an organization, but allowing other organizations to update your users seems sketchy, unless this is the only real update we're allowing.

datastore/migrations/0012_role_management.up.sql
```diff
@@ -0,0 +1,398 @@
...
+    SET rid = UUID_TO_BIN(roleid, 1);
+    SET uid = UUID_TO_BIN(userid, 1);
+    SET allowed = can_user_perform_action(auth0id, uid, 'update');
+    IF allowed THEN
+        DELETE FROM arbiter_data.user_role_mapping WHERE user_id = uid AND role_id = rid;
+    ELSE
+        SIGNAL SQLSTATE '42000' SET MESSAGE_TEXT = 'Access denied to user on "remove role from user"',
+        MYSQL_ERRNO = 1142;
+    END IF;
+END;
```
lboeman: Here as well.

datastore/migrations/0012_role_management.up.sql
alorenzo175: good point. I should actually be more careful in other functions as well. We don't want someone creating permissions to objects that the org doesn't own.

alorenzo175:
I've changed it to require that adding things (users, roles, permissions) requires the user adding, and the two objects (user + role, role+permission, permission+object) all have the same organization. For removing the user performing the remove and the object being updated must have the same org. I think this will prevent adding permission to objects outside of the organization.

Sharing roles with users in other organizations will be done separately and will require user consent.

datastore/tests/test_authorization.py
```diff
@@ -210,3 +216,208 @@ def test_user_can_create_multiple_permissions(cursor, make_user_roles,
...
+                                        new_permission):
+    action = 'read'
+    otype = 'observations'
+    auth0id, org = rbac_user(otype, False)
+    others = make_test_permissions(org)
+    perm = new_permission(action, otype, False, org=org)
+    if otype == 'users':
+        oid = others['user']['id']
+    else:
+        oid = others[otype][0]['id']
```
lboeman: I think this conditional may be leftover copypasta from above where otype is parametrized? also in the following tests.

datastore/tests/test_authorization.py
```diff
@@ -210,3 +216,208 @@ def test_user_can_create_multiple_permissions(cursor, make_user_roles,
...
+    action = 'create'
+    otype = 'forecasts'
+    _, org = rbac_user(otype)
+    auth0id = new_user()['auth0_id']
+    others = make_test_permissions(org)
+    perm = new_permission(action, otype, False, org=org)
+    if otype == 'users':
+        oid = others['user']['id']
+    else:
+        oid = others[otype][0]['id']
```
lboeman: here

datastore/tests/test_authorization.py
```diff
@@ -210,3 +216,208 @@ def test_user_can_create_multiple_permissions(cursor, make_user_roles,
...
+
+    action = 'read'
+    otype = 'sites'
+    auth0id, org = rbac_user(otype)
+    others = make_test_permissions(org)
+    perm = new_permission(action, otype, False)
+    if otype == 'users':
+        oid = others['user']['id']
+    else:
+        oid = others[otype][0]['id']
```
lboeman: here

datastore/tests/test_authorization.py
```diff
@@ -210,3 +216,208 @@ def test_user_can_create_multiple_permissions(cursor, make_user_roles,
...
+                                            new_organization):
+    action = 'delete'
+    otype = 'roles'
+    auth0id, org = rbac_user(otype)
+    others = make_test_permissions(new_organization())
+    perm = new_permission(action, otype, False, org=org)
+    if otype == 'users':
+        oid = others['user']['id']
+    else:
+        oid = others[otype][0]['id']
```
lboeman: also here

Issue #83, opened 2019-04-04
Deletion 404s


lboeman:
Since mysql fails silently when something doesn't exist, we should check for this and return something. Currently, deleting an object or trying to delete a missing object returns a 200 and trying to delete something you don't have access to throws a 404.

alorenzo175:
I don't quite understand. Seems to me that only deleting a valid object that the user has permission to delete returns 204. Otherwise you get a 404. Do you mean if a user has read access to the object but not delete, trying to delete the object should return something else?

lboeman:
No it looks like you're right, perhaps I had something broken in my branch. I had added tests to check that attempting to delete non existent objects returned a 404, and they were breaking with a 200 status code. I'll add those tests back in.

PR #109, opened 2019-05-08
Add immediate validation on observations post

alorenzo175:
wait for https://github.com/SolarArbiter/solarforecastarbiter-core/pull/83

### DIFF
--- a/sfa_api/utils/request_handling.py
+++ b/sfa_api/utils/request_handling.py
@@ -69,7 +69,8 @@ def validate_observation_values(observation_df, quality_flag_range=(0, 1)):
         if not observation_df['quality_flag'].between(
                 *quality_flag_range).all():
             errors['quality_flag'].append(
-                'Item in "quality_flag" field out of range.')
+                'Item in "quality_flag" field out of range '
+                f'{quality_flag_range}.')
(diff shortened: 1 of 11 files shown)


sfa_api/utils/request_handling.py
```diff
@@ -1,17 +1,87 @@
...
+    except KeyError:
+        errors['quality_flag'].append('Missing "quality_flag" field.')
+    except (ValueError, TypeError):
+        errors['quality_flag'].append(
+            'Item in "quality_flag" field is not an integer.')
+    else:
+        if not np.isclose(
+                observation_df['quality_flag'].mod(1), 0, 1e-12).all():
+            errors['quality_flag'].append(
+                'Item in "quality_flag" field is not an integer.')
```
lboeman: It seems like this `if not np.isclose()` could probably replace the `astype(int)` call above to verify the columns existence and integer-ness. I hadn't realized that the casting done by astype will floor floats.

alorenzo175: I had that originally, but if there are any strings or things in the series, .mod is unhappy. Could catch that error, but I think it's better to be explicit

PR #111, opened 2019-05-09
add reference user and role for accessing all reference data

lboeman:
Adds the reference data user, with role to create, read, and delete all reference data.


lboeman:
> Does the reference user really need to be able to delete things? I know you have a delete command in the script, but I don't think that is necessarily a good idea either.

I agree that not allowing it in Production would be a positive thing. But it would be a pain to develop without it. Unless you're thinking that we'd have an 'Admin' user with access to everything.  

Maybe we should move all of the demo/ development oriented migrations later so that we can initialize a clean production database by running migrations up to some number, but then initialize with Dev by running migrations past that?

PR #115, opened 2019-05-11
use RQ

alorenzo175:
after solararbiter/solarforecastarbiter-core#86
Move queuing to the API and use RQ as the queue. RQ is even simpler than dramatiq now that we've decided redis is an OK message broker (and used in production by the likes of Netflix). It also better decouples the queuing from the actual tasks.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,3 +19,4 @@ redis
 rq
 fakeredis
 sentry_sdk
+blinker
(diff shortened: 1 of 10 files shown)


alorenzo175:
You mean the validation task in ``observations.py``? I think the answer is no because none of the tasks I'm thinking of now will be returned in a single request. With a working Redis connection, the tasks should all be queued up and run after the original request has already returned. Other tasks include longer term validation and generating reports.

lboeman:
Yeah, I just misunderstood what was happening there. I didn't notice the validator function posting results at first. Good to merge with me.

PR #120, opened 2019-05-20
Role/Permission Endpoints

lboeman:
Adds User, Role and Permissions endpoints to the API. User creation/deletion endpoints are just method stubs and documentation at the moment to get feedback on the layout of the endpoints, I noted the procedures that each endpoint would end up calling.
I added the User->Role mapping functions to Users, but tagged them with both 'Users' and 'Roles'  so that they can be found in both sections of the documentation, this could be repeated for the Roles -> Permissions mappings.

A couple of things I noted while working on this:
- There is no `delete_user` procedure, should that exist or is that going to be something we defer to auth0?
- Should auth0_id be treated as a secret, and the internal SolarForecastArbiter UUID used to identify users? other way around? neither?

### DIFF
--- a/sfa_api/conftest.py
+++ b/sfa_api/conftest.py
@@ -144,8 +144,8 @@ def _make_sql_app():
             yield app
 
 
-@pytest.fixture()
-def sql_app(mocker):
+@pytest.fixture(scope='module')
+def sql_app():
     with _make_sql_app() as app:
         yield app
 
--- a/sfa_api/conftest.py
+++ b/sfa_api/conftest.py
@@ -154,8 +154,7 @@ def sql_app(mocker):
 def sql_app_no_commit(mocker):
     with _make_sql_app() as app:
         with _make_nocommit_cursor(mocker):
-                yield app
-
+            yield app
 
 
 @pytest.fixture()
--- a/sfa_api/conftest.py
+++ b/sfa_api/conftest.py
@@ -144,10 +144,18 @@ def _make_sql_app():
             yield app
 
 
-@pytest.fixture(scope='module')
-def sql_app():
+@pytest.fixture()
+def sql_app(mocker):
     with _make_sql_app() as app:
         yield app
+
+
+@pytest.fixture()
+def sql_app_no_commit(mocker):
+    with _make_sql_app() as app:
+        with _make_nocommit_cursor(mocker):
+                yield app
+
 
 
 @pytest.fixture()
--- a/sfa_api/schema.py
+++ b/sfa_api/schema.py
@@ -29,7 +29,10 @@ VARIABLE_FIELD = ma.String(
     description="The variable being forecast",
     required=True,
     validate=validate.OneOf(VARIABLES))
-
+ORGANIZATION_ID = ma.UUID(
+        title="Organization ID",
+        description="UUID of the Organization the Object belongs to."
+    )
 CREATED_AT = ma.DateTime(
     title="Creation time",
     description="ISO 8601 Datetime when object was created",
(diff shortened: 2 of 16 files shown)


sfa_api/schema.py
```diff
@@ -526,3 +526,39 @@ class CDFForecastGroupSchema(CDFForecastGroupPostSchema):
...
+@spec.define_schema('UserPostSchema')
+class UserPostSchema(ma.Schema):
+    auth0_id = ma.String(
+        title="Auth0 ID",
+        description="The User's unique Auth0 identifier.",
+    )
+    # Not sure if this should be here, on one hand we want Admin at an
+    # Organization to only be able to add Users for their organization,
+    # but framework administrators would need a way to add a user to any
+    # organization.
```
alorenzo175: For now, I would leave off user creation/deletion. They will require integration with Auth0 and I'm not sure how that will work yet.

sfa_api/schema.py
```diff
@@ -526,3 +526,39 @@ class CDFForecastGroupSchema(CDFForecastGroupPostSchema):
...
+@spec.define_schema('UserSchema')
+class UserSchema(ma.Schema):
+    _links = ma.Hyperlinks(
+        {
+            'roles': ma.AbsoluteURLFor('roles.user_roles',
+                                       user_id='<user_id>'),
+        },
+        description="Contains a link to the Users associated Roles."
+    )
+    # Should auth0_id be included in the response schema?
```
alorenzo175: probably just uuid

sfa_api/schema.py
```diff
@@ -526,3 +526,39 @@ class CDFForecastGroupSchema(CDFForecastGroupPostSchema):
...
+    user_id = ma.UUID(
+        title="User ID",
+        description="Unique UUID of the User.",
+    )
+    organization_id = ma.UUID(
+        title="Organization ID",
+        description="UUID of the Organization the User belongs to."
+    )
```
alorenzo175: not sure if this will have more information retrieved from auth0

lboeman:
@alorenzo175  While working on this, I realized the users/roles/permissions procedures are defined with other MYSQL users. Is the MYSQL user 'apiuser' meant to have grants to execute the CRUD procedures for those types of objects? I added those permissions locally for testing by hand while working on this, but I wasn't sure if there is another plan for that?

alorenzo175:
Yes, apiuser should have the execute privileges on the functions the python code uses

lboeman:
Opened a couple of further issues I found while working on this #122 and #123 . They are not breaking issues, but are unexpected behavior.

sfa_api/conftest.py
```diff
@@ -144,12 +144,19 @@ def _make_sql_app():
             yield app
 
 
-@pytest.fixture(scope='module')
-def sql_app():
+@pytest.fixture()
+def sql_app(mocker):
```
alorenzo175: why change this? maybe leftover from when you where trying to get permissions to work properly

sfa_api/schema.py
```diff
@@ -526,3 +529,98 @@ class CDFForecastGroupSchema(CDFForecastGroupPostSchema):
...
+        validate=validate.OneOf(['create', 'read', 'update',
+                                 'delete', 'read_values', 'write_values',
+                                 'delete_values']),
+    )
+    object_type = ma.String(
+        title="Object Type",
+        description="The type of object this permission will act on.",
+        required=True,
+        validate=validate.OneOf(['sites', 'aggregates', 'forecasts',
+                                 'observations', 'users', 'roles',
```
alorenzo175: need cdf_forecasts

PR #124, opened 2019-05-30
Add data access example users

lboeman:
Sets up three users for demonstrating data access controls. One utility with site and observations, and two Forecasters with forecasts at the utility's site as well as Ashland OR. The utility can read it's own data and both of the forecasts at it's own site, but cannot read the Ashland OR site or the forecasts made there. Each Forecaster can read their own forecasts and the utility's site as well as Ashland OR but cannot see eachothers forecasts. 
The users can be accessed with the following credentials:
forecastera@solarforecastarbiter.org:Forecasterpassword!
forecasterb@solarforecastarbiter.org:Forecasterpassword!
utilityx@solarforecastarbiter.org:Utilitypassword!

### DIFF
--- a/datastore/migrations/0022_permissions_demo.down.sql
+++ b/datastore/migrations/0022_permissions_demo.down.sql
@@ -1,7 +1,7 @@
 DELETE FROM arbiter_data.organizations WHERE name = 'Forecast Provider A';
 DELETE FROM arbiter_data.organizations WHERE name = 'Forecast Provider B';
 DELETE FROM arbiter_data.organizations WHERE name = 'Utility X';
-DELETE FROM arbiter_data.roles WHERE name = 'Read Ashland Site';
-DELETE FROM arbiter_data.permissions WHERE description = 'Read Ashland OR Site';
-DELETE FROM arbiter_data.permissions WHERE description = 'Read Ashland OR Observations';
-DELETE FROM arbiter_data.permissions WHERE description = 'Read Ashland OR Observation values';
+DELETE FROM arbiter_data.roles WHERE name = 'Read Weather Station Site';
+DELETE FROM arbiter_data.permissions WHERE description = 'Read Weather Station Site';
+DELETE FROM arbiter_data.permissions WHERE description = 'Read Weather Station Observations';
+DELETE FROM arbiter_data.permissions WHERE description = 'Read Weather Station Observation values';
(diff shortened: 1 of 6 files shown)


datastore/migrations/0022_permissions_demo.up.sql
```diff
@@ -0,0 +1,328 @@
...
+-- Create a Power plant for Utilitty X 
+SET @plant_id = (UUID_TO_BIN('c3f01422-824d-11e9-a81f-54bf64606445', 1));
+INSERT INTO arbiter_data.sites (
+    id, organization_id, name, latitude, longitude, elevation, timezone, extra_parameters,
+    ac_capacity, dc_capacity, temperature_coefficient, tracking_type, surface_tilt, surface_azimuth,
+    ac_loss_factor, dc_loss_factor, created_at, modified_at
+) VALUES (
+    @plant_id, 
+    @utilxorgid, 
+    'Power Plant X', 43.73403, -96.62328, 786.0, 'Etc/GMT+6', '', 0.015, 0.015,
```
wholmgren: nitpick for realism: 10 MW AC, 12 MW DC

lboeman:
> is there any point in inserting data as well? I don't recall if the current implementation supports sharing of data separately from metadata

Reading values is a separate permission from reading metadata. Though I'm not sure what the actual use case is for allowing a user permission to one and not the other.

wholmgren:
> Though I'm not sure what the actual use case is for allowing a user permission to one and not the other.

Some people are particular about their data. For the exercise for next week, let's write that it's possible to do that but don't implement a demonstration.

lboeman:
Updated this to add the ability to access reference data to all three test users, I'll work on adding an Organization filter to fx/obs/site listing pages next.

lboeman:
I think this is ready to go now. The three users now have access to reference data and I've updated the 'Ashland OR' to 'Weather Station' to avoid confusion wiith the actual Ashland OR site from SRML. I'll update  the Denver Workshop text to reflect filtering out Reference data from pages. I've opened https://github.com/SolarArbiter/solarforecastarbiter_dashboard/pull/84 to add a filter to the data listing pages.

datastore/migrations/0022_permissions_demo.down.sql
```diff
@@ -0,0 +1,7 @@
+DELETE FROM arbiter_data.organizations WHERE name = 'Forecast Provider A';
+DELETE FROM arbiter_data.organizations WHERE name = 'Forecast Provider B';
+DELETE FROM arbiter_data.organizations WHERE name = 'Utility X';
+DELETE FROM arbiter_data.roles WHERE name = 'Read Ashland Site';
+DELETE FROM arbiter_data.permissions WHERE description = 'Read Ashland OR Site';
+DELETE FROM arbiter_data.permissions WHERE description = 'Read Ashland OR Observations';
+DELETE FROM arbiter_data.permissions WHERE description = 'Read Ashland OR Observation values';
```
wholmgren: do these descriptions need to be updated?

lboeman: Yep. good catch. thanks.

Issue #20, opened 2018-11-08
draft of /reports blueprint


alorenzo175:

lboeman:
I'm taking a stab at adding the MYSQL tables to make reports go. To start, the list of metrics are a JSON array, and data is a blob.

PR #132, opened 2019-06-10
draft reports endpoints

lboeman:
Working towards #20.
Adds method stubs for reports endpoints.


lboeman:
@alorenzo175 It is my understanding that we're going to have the API return Reports as an object including any metadata, parameters, calculated metrics and processed/aligned data. If we're storing all of this data, there is a case for permissions we haven't covered.
If a user has permissions to read a report and does not have access to read the data used to generate it, should that user be able to view a summary report of the metrics without the included time series plots of the original data? Or should they be completely blocked from viewing the report?

alorenzo175:
> If a user has permissions to read a report and does not have access to read the data used to generate it, should that user be able to view a summary report of the metrics without the included time series plots of the original data?

yes. This is part of the reason for separating the data by object_id in report_values. The idea is that only values that the user has read access for object_id are returned. Right now, I think this means the api will call two procedures. The first will be to get the report with the metadata, metrics, etc from the reports table. The second will be to retrieve any data for that report that the user has read_values permissions for the corresponding object_id.

lboeman:
@alorenzo175 can you take a look at the reports migration here when you have a chance? I think it's in good shape, but not sure if we should have a manual update procedure for setting status, or if it should be triggered when metrics are inserted.

datastore/migrations/0013_reports_tables.up.sql
```diff
@@ -0,0 +1,197 @@
...
+    FOREIGN KEY (report_id)
+        REFERENCES arbiter_data.reports(id)
+        ON DELETE CASCADE ON UPDATE RESTRICT
+) ENGINE=INNODB ENCRYPTION='Y' ROW_FORMAT=COMPRESSED;
+
+-- Report READ PROCEDURES
+CREATE DEFINER = 'select_objects'@'localhost' PROCEDURE list_reports (IN auth0id VARCHAR(32))
+READS SQL DATA SQL SECURITY DEFINER
+SELECT BIN_TO_UUID(id, 1) as report_id, get_organization_name(organization_id) as provider,
+       name, report_parameters, metrics
```
alorenzo175: why not the other columns too? status, created_at, modified_at

datastore/migrations/0013_reports_tables.up.sql
```diff
@@ -0,0 +1,197 @@
...
+COMMENT 'Read report metadata'
+READS SQL DATA SQL SECURITY DEFINER
+BEGIN
+    DECLARE binid BINARY(16);
+    DECLARE allowed BOOLEAN DEFAULT FALSE;
+    SET binid = (SELECT UUID_TO_BIN(strid, 1));
+    SET allowed = (SELECT can_user_perform_action(auth0id, binid, 'read'));
+    IF allowed THEN
+        SELECT BIN_TO_UUID(id, 1) as report_id, get_organization_name(organization_id) as provider,
+        name, report_parameters, metrics
```
alorenzo175: same as above, what about other columns

datastore/migrations/0013_reports_tables.up.sql
```diff
@@ -0,0 +1,197 @@
...
+
+    PRIMARY KEY (id),
+    FOREIGN KEY (organization_id)
+        REFERENCES arbiter_data.organizations(id)
+        ON DELETE CASCADE ON UPDATE RESTRICT
+) ENGINE=INNODB ENCRYPTION='Y' ROW_FORMAT=COMPRESSED;
+
+-- create table for storing reports values
+CREATE TABLE arbiter_data.report_values(
+    report_id BINARY(16) NOT NULL,
```
alorenzo175: I think this needs an additonal ``id`` column which can be used by the report to determine what the process_values are. There might be cases where the same object is processed in multiple ways for the report, especially when comparing multiple forecasts that might have different interval length/averaging.

datastore/migrations/0013_reports_tables.up.sql
```diff
@@ -0,0 +1,197 @@
...
+    object_id BINARY(16) NOT NULL,
+    processed_values BLOB NOT NULL,
+    
+    PRIMARY KEY (report_id, object_id),
```
alorenzo175: Adding the id also means this primary key wont work since (report_id, object_id) aren't necessarily unique pairs. Furthermore, we need a trigger that deletes the processed data if the forecast/observation it is derived from is deleted and this ordering of the primary key is not optimal for those object_id lookups. So, I think the primary key should be (report_id, id) and object_id should just be a KEY for quick lookups on delete.

alorenzo175: actually might need to double check how the foreign key index works. If it always makes a new index, then the primary key should just be (id). If it can reuse the primary key index, (report_id, id) as the primary key should work

datastore/migrations/0013_reports_tables.up.sql
```diff
@@ -0,0 +1,197 @@
...
+MODIFIES SQL DATA SQL SECURITY DEFINER
+COMMENT 'Store processed values for a report'
+BEGIN
+    DECLARE bin_report_id BINARY(16);
+    DECLARE bin_object_id BINARY(16);
+    DECLARE read_allowed BOOLEAN DEFAULT FALSE;
+    DECLARE create_allowed BOOLEAN DEFAULT FALSE;
+    SET bin_report_id = (SELECT UUID_TO_BIN(str_report_id, 1));
+    SET bin_object_id = (SELECT UUID_TO_BIN(str_object_id, 1));
+    SET read_allowed = (SELECT can_user_perform_action(auth0id, bin_object_id, 'read_values'));
```
alorenzo175: I don't think this is necessary to check here.

datastore/migrations/0013_reports_tables.up.sql
```diff
@@ -0,0 +1,197 @@
...
+CREATE DEFINER = 'permission_trig'@'localhost' TRIGGER add_object_perm_on_reports_insert AFTER INSERT ON arbiter_data.reports
+FOR EACH ROW INSERT INTO arbiter_data.permission_object_mapping (permission_id, object_id)
+    SELECT id, NEW.id FROM arbiter_data.permissions WHERE action != 'create' AND applies_to_all AND organization_id = NEW.organization_id AND object_type = 'reports';
+
+
+-- remove any object permission mappings after report deletion
+CREATE DEFINER = 'permission_trig'@'localhost' TRIGGER remove_object_perm_on_report_delete AFTER DELETE ON arbiter_data.reports
+FOR EACH ROW DELETE FROM arbiter_data.permission_object_mapping WHERE object_id = OLD.id;
+
+
```
alorenzo175: Need the triggers to delete from report_values if the object_id is deleted from forecasts/observations/cdf forecasts

lboeman:
Updated the endpoints/schema:
Users will POST to:
**/reports/** for adding the report parameters
**/reports/<report_id>/values** for storing processed data, getting a uuid back
**/reports/<report_id>/metrics** for storing metrics
**/reports/<report_id>/<status>** for updating report status, where status is one of: pending, failed, complete

Users can GET:
**/reports/** For a list of reports, these objects exclude values
**/reports/<report_id>** for individual report metadata including values
**/reports/<report_id>/values ** For a report's values. I'm not sure this is necessary. can be removed.

sfa_api/reports.py
```diff
@@ -0,0 +1,261 @@
...
+            description: Successfully retrieved
+            content:
+              application/json:
+                schema:
+                  $ref: '#/components/schemas/ReportValuesSchema'
+          404:
+            $ref: '#/components/responses/404-NotFound'
+        """
+        # Maybe this shouldn't exist since we're packing the values
+        # into the single report
```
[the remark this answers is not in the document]
alorenzo175: not sure how all of this will work though with the result set from sql

datastore/migrations/0013_reports_tables.up.sql
```diff
@@ -0,0 +1,302 @@
+ALTER TABLE arbiter_data.permissions CHANGE COLUMN object_type object_type ENUM('sites', 'aggregates', 'cdf_forecasts', 'forecasts', 'observations', 'users', 'roles', 'permissions', 'reports') NOT NULL;
+
+-- create reports table
+CREATE TABLE arbiter_data.reports(
+    id BINARY(16) NOT NULL DEFAULT (UUID_TO_BIN(UUID(), 1)),
+    organization_id BINARY(16) NOT NULL,
+    name VARCHAR(64) NOT NULL,
+    report_parameters JSON NOT NULL,
+    metrics JSON,
+    status ENUM('pending', 'complete', 'failed') NOT NULL,
```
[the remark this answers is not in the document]
lboeman: I searched for how to explicitly set this, but mysql documentation just says this: 

>  If an ENUM column is declared NOT NULL, its default value is the first element of the list of permitted values
> from https://dev.mysql.com/doc/refman/8.0/en/enum.html#enum-nulls

sfa_api/reports.py
```diff
@@ -0,0 +1,261 @@
...
+reports_blp.add_url_rule(
+    '/',
+    view_func=AllReportsView.as_view('all')
+)
+reports_blp.add_url_rule(
+    '/<report_id>',
+    view_func=ReportView.as_view('single')
+)
+reports_blp.add_url_rule(
+    '/<report_id>/status/<status>',
```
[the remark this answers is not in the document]
lboeman: I set status as a path parameter here. It seems easier to just run `POST /<uuid>/status/complete` to set the status since there are only 3 predefined states. Perhaps for consistency it should use a request body like the rest of the API though?

PR #138, opened 2019-07-23
handle file uploads

### LINKED ISSUE #130: handle excel files
of content-type 'application/vnd.ms-excel'. Perhaps first assume they are csv and if parsing fails, raise a good error. Mainly for dashboard https://github.com/SolarArbiter/solarforecastarbiter_dashboard/pull/101

lboeman:
Need to figure out why one of the tests is still failing, and possibly put the tests somewhere else. I've added a check for failure to decode as UTF-8 as in my testing, that is the error from trying to upload a non-csv application/vnd.ms-excel file. 
closes #138 
closes #130

### DIFF
--- a/sfa_api/tests/test_observations.py
+++ b/sfa_api/tests/test_observations.py
@@ -312,5 +312,5 @@ def test_post_file_invalid_mimetype(api, observation_id):
         content_type='multipart/form-data',
         data=incorrect_file_payload)
     assert file_post.status_code == 400
-    expected = '{"errors":{"error":["Uploaded file has invalid mimetype."]}}\n'
+    expected = '{"errors":{"error":["Unsupported Content-Type or MIME type."]}}\n' # noqa
     assert file_post.get_data(as_text=True) == expected
--- a/sfa_api/tests/test_observations.py
+++ b/sfa_api/tests/test_observations.py
@@ -226,14 +226,17 @@ def dummy_file():
     return req_file
 
 
-@pytest.mark.parametrize('filename,str_content,content_type', [
-    ('data.csv', VALID_OBS_VALUE_CSV, 'text/csv'),
-    ('data.csv', VALID_OBS_VALUE_CSV, 'application/vnd.ms-excel'),
-    ('data.json', json.dumps(VALID_OBS_VALUE_JSON), 'application/json'),
+@pytest.mark.parametrize('filename,str_content,content_type,start,end', [
+    ('data.csv', VALID_OBS_VALUE_CSV, 'text/csv',
+     '2019-01-22T12:04:00+00:00', '2019-01-22T12:07:00+00:00'),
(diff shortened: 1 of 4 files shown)


lboeman:
Pandas 0.25.0 introduces a change when slicing with two datetimes with different UTC offsets throws and error, so I've added a helper function to convert timestamps to utc before slicing.

sfa_api/utils/request_handling.py
```diff
@@ -85,28 +143,44 @@ def validate_parsable_values():
...
     """
-    if request.content_type == 'application/json':
-        raw_data = request.get_json()
+    if request.mimetype == 'multipart/form-data':
+        # handle single file uploads
         try:
-            raw_values = raw_data['values']
-        except (TypeError, KeyError):
-            error = 'Supplied JSON does not contain "values" field.'
+            posted_filename = list(request.files.keys())[0]
```
alorenzo175: what if someone tries to upload multiple files? I think we should send an error back if we're not supporting that

sfa_api/utils/request_handling.py
```diff
@@ -85,28 +143,44 @@ def validate_parsable_values():
...
-                                   comment='#')
-        except pd.errors.EmptyDataError:
-            raise BadAPIRequest({'error': 'Malformed CSV'})
-        finally:
-            raw_data.close()
+        if (
+                posted_file.mimetype == 'text/csv' or
+                posted_file.mimetype == 'application/vnd.ms-excel' or
+                posted_file.mimetype == 'application/json'
+        ):
```
alorenzo175: instead of multiple ``if posted_file.mimetype == ...`` here and below, put the contents of this if statement in a function and then just have the unique ``if posted_file.mimetype ==`` below that call this new function for ``decoded_data``

sfa_api/tests/test_observations.py
```diff
@@ -215,3 +217,100 @@ def test_post_and_get_values_csv(api, observation_id, mocked_queuing):
...
+        'data.csv': (BytesIO(b'invalid'), 'data.xls', 'application/videogame')
+    }
+    file_post = api.post(
+        f'/observations/{observation_id}/values',
+        base_url=BASE_URL,
+        content_type='multipart/form-data',
+        data=incorrect_file_payload)
+    assert file_post.status_code == 400
+    expected = '{"errors":{"error":["Uploaded file has invalid mimetype."]}}\n'
+    assert file_post.get_data(as_text=True) == expected
```
alorenzo175: probably not a bad idea to add tests for parse_csv and parse_json separately from the full post

lboeman:
I believe sprucing has been achieved where need of sprucing was noted. I think the refactor to the values parsing is easier to read and achieves better separation of concerns. Hopefully it's still in line with what you suggested, if not happy to take another look.

PR #139, opened 2019-07-24
Sentry issues

lboeman:
Aims to Close a handful of issues found with Sentry. The tests will not pass until the changes in #130 are merged due to the issue with pandas 0.25.0

- Adds validation handling for 'NaT' as a query parameter. 
- Handles the empty list returned when a user has permission to read a uuid of one type, but tries to access it from another endpoint.
- Enforces lowercase metrics for compatibility with core.

### DIFF
--- a/sfa_api/utils/request_handling.py
+++ b/sfa_api/utils/request_handling.py
@@ -228,6 +228,28 @@ def validate_parsable_values():
     return value_df
 
 
+def parse_to_timestamp(dt_string):
+    """Attempts to parse to Timestamp.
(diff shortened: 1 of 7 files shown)


sfa_api/utils/request_handling.py
```diff
@@ -228,6 +228,28 @@ def validate_parsable_values():
     return value_df
 
 
+def parse_to_timestamp(dt_string):
```
alorenzo175: should test this separately too

Issue #144, opened 2019-08-19
Foreign key contraints on shared site prohibit organization deletion


lboeman:
While working on data sharing updates I discovered an issue with organization deletion
Steps to reproduce:
- Organization A grants read access to a site for some user in Organization B
- A user from another organization submits a forecast for that site.
- Attempt to delete Organization A fails due to the Foreign Key constraint created by the forecast added by Organization B.

 It seems that an Organization should always be able to remove themselves from the framework. With that in mind, here are some possible solutions:

- When deleting a site, delete all observations and forecasts related to that site, regardless of owner. We'd need to establish some End User documentation that clearly defines the permission implications of submitting data to a site owned by another organization. It does seem difficult to reflect the ability to delete these objects with permissions, e.g. might need a new role and permission need to be made for the Site's organization for Each provided forecast/observation.
- If other organizations have forecasts or observations for the site, create a copy of that site, perhaps under the 'Unaffiliated' organization.
- Orphan the Observations and Forecasts from other organizations. It seems like orphaned observations and forecasts are otherwise useless without the metadata provided by their associated site.

alorenzo175:
I think I'm OK with just orphaning any forecast/obs. Any new thoughts @lboeman?

lboeman:
That sounds fine to me. I think protecting against deleting an in-use site is probably worth while, either through a check at the API or just with a confirmation message on the dashboard. I think maybe assuming someone programming against the API would know to check is fine.

alorenzo175:
I guess the problem with checking if a site is in use is that the person deleting said site may not have permission to view whatever is using it. So maybe we should just make the warning message when deleting the site in the dashboard say that other forecasts/obs will be orphaned

lboeman:
Ahh good point. I agree.

PR #145, opened 2019-08-19
Data Sharing and User management updates

lboeman:
Updates for sharing data and user management. Still a work in progress.
Updates as of this moment:  
- Updated adding/removing roles from users to depend on the 'create role grants' permission.
- Added check for rbac permissions on a role before sharing with users outside the callers org.
- Added check for role granted to external user before allowing adding rbac-based permissions to a role.
- Create a new user if they have a valid auth0id and their email is validated. 
- Add a procedure and endpoint for listing the current user's information (uuid, org_id, org name)
- Add a procedure/endpoint for listing users that are granted roles from your organization. Again, the 'create role_grants' permission is checked here. 

Introduces #144 

I've used the create 'role_grants' permission as a sort of admin flag here, Users could be allowed to remove roles from themselves without a permissions check by comparing their auth0id with the mapping. We could alternatively apply a uuid to each role_user mapping, and have permissions to control the crud operations on each mapping. Allowing users to remove themselves from the role would be complicated though, as a permission to delete the specific user_role mapping would make the role user specific.


datastore/migrations/0023_data_sharing.up.sql
```diff
@@ -0,0 +1,289 @@
...
+
+CREATE DEFINER = 'select_rbac'@'localhost' FUNCTION role_contains_rbac_permissions(roleid BINARY(16))
+RETURNS BOOLEAN
+COMMENT 'Determines if a role contains rbac object permissions'
+READS SQL DATA SQL SECURITY DEFINER
+BEGIN
+    RETURN (SELECT EXISTS (
+        SELECT 1 FROM arbiter_data.role_permission_mapping
+            WHERE role_id = roleid
+            AND rbac_permissions_check(permission_id)
```
alorenzo175: I've forgotten exactly how this works in mysql, but we definitely need to test this with multiple permissions for a role including roles with some rbac permissions and some other permissions

datastore/migrations/0023_data_sharing.up.sql
```diff
@@ -0,0 +1,289 @@
...
+    DECLARE userid BINARY(16);
+    DECLARE role_org BINARY(16);
+    DECLARE caller_org BINARY(16);
+    DECLARE grantee_org BINARY(16);
+
+    SET roleid = UUID_TO_BIN(role_id, 1);
+    SET role_org = get_object_organization(roleid, 'roles');
+    SET caller_org = get_user_organization(auth0id);
+    SET userid = UUID_TO_BIN(user_id, 1);
+    SET grantee_org = (SELECT organization_id FROM arbiter_data.users WHERE id = userid);
```
alorenzo175: If no current procedure works for this, make a new one instead of granting the SELECT permission

alorenzo175: Maybe this function returns Null if the user doesn't exist and then that null check can be used to check if allowed

datastore/migrations/0023_data_sharing.up.sql
```diff
@@ -0,0 +1,289 @@
...
+    DECLARE role_org BINARY(16);
+    DECLARE userid BINARY(16);
+    DECLARE caller_org BINARY(16);
+    SET caller_org = get_user_organization(auth0id);
+    SET roleid = UUID_TO_BIN(role_strid, 1);
+    SET role_org = get_object_organization(roleid, 'roles');
+    SET userid = UUID_TO_BIN(user_strid, 1);
+    -- calling user must have create role_grants
+    -- calling user and role must be in the same organization
+    SET allowed = user_can_create(auth0id, 'role_grants') AND
```
alorenzo175: should check if user can delete role_grants

datastore/migrations/0023_data_sharing.up.sql
```diff
@@ -0,0 +1,289 @@
...
+CREATE DEFINER = 'insert_rbac'@'localhost' PROCEDURE create_user_if_not_exists(IN auth0id VARCHAR(32))
+COMMENT 'Inserts a new user and adds then to the Public org, and read reference role'
+READS SQL DATA SQL SECURITY DEFINER
+BEGIN
+    DECLARE orgid BINARY(16);
+    DECLARE userid BINARY(16);
+    DECLARE reference_roleid BINARY(16);
+    SET userid = UUID_TO_BIN(UUID(), 1);
+    SET orgid = (SELECT id FROM arbiter_data.organizations WHERE name = "Unaffiliated");
+    SET reference_roleid = (SELECT id FROM arbiter_data.roles WHERE name = 'Read Reference Data');
```
alorenzo175: what if someone else decides to name a role 'Read Reference Data'?

should also be a function

datastore/migrations/0023_data_sharing.up.sql
```diff
@@ -0,0 +1,289 @@
...
+ */
+CREATE DEFINER = 'insert_rbac'@'localhost' PROCEDURE create_user_if_not_exists(IN auth0id VARCHAR(32))
+COMMENT 'Inserts a new user and adds then to the Public org, and read reference role'
+READS SQL DATA SQL SECURITY DEFINER
+BEGIN
+    DECLARE orgid BINARY(16);
+    DECLARE userid BINARY(16);
+    DECLARE reference_roleid BINARY(16);
+    SET userid = UUID_TO_BIN(UUID(), 1);
+    SET orgid = (SELECT id FROM arbiter_data.organizations WHERE name = "Unaffiliated");
```
alorenzo175: should be a function and avoid granting mysql directly on tables

datastore/migrations/0023_data_sharing.up.sql
```diff
@@ -0,0 +1,289 @@
...
+    FROM arbiter_data.users WHERE auth0_id = auth0id;
+END;
+GRANT EXECUTE ON PROCEDURE arbiter_data.get_current_user_info TO 'select_rbac'@'localhost';
+GRANT EXECUTE ON FUNCTION arbiter_data.get_organization_name TO 'select_rbac'@'localhost';
+GRANT EXECUTE ON PROCEDURE arbiter_data.get_current_user_info TO 'apiuser'@'%';
+
+/*
+ * List users with role in the organization.
+ */
+CREATE DEFINER = 'select_rbac'@'localhost' PROCEDURE list_priveleged_users (
```
alorenzo175: typo:  privileged

datastore/migrations/0023_data_sharing.up.sql
```diff
@@ -0,0 +1,289 @@
...
+    IN auth0id VARCHAR(32))
+COMMENT 'List all users who are granted a role within the organization'
+BEGIN
+    DECLARE orgid BINARY(16);
+    DECLARE allowed BOOLEAN;
+    SET allowed = user_can_create(auth0id, 'role_grants');
```
alorenzo175: wouldn't this more appropriately be user can read?

datastore/migrations/0023_data_sharing.up.sql
```diff
@@ -0,0 +1,289 @@
...
+    IF allowed THEN
+        SET orgid = get_user_organization(auth0id);
+        SELECT BIN_TO_UUID(id, 1) as user_id, get_organization_name(organization_id) as organization, BIN_TO_UUID(organization_id) as organization_id, auth0_id
```
alorenzo175: why expose auth0_id at all? seems like only the user_id is needed for the admin actions

alorenzo175: Why not also list the roles that are assigned to the users?

datastore/migrations/0023_data_sharing.up.sql
lboeman: I've use create here, because user_role_mapping entries do not have associated UUIDS. Currently all of the permissions besides 'create' require individual object mappings, should I update the user_role mappings with UUIDS so this check fits into the current RBAC structure?

alorenzo175: Ah, I think the problem then is that role_grants should not be an object at all. I think it makes more sense to add grant and revoke actions to permissions instead of the role_grants object_type.

lboeman:
@alorenzo175  I think I've touched upon all of your comments. I'm going to need to fix the tests, which is failing in part due to the demo module not containing new storage_interface functions.  To implement the the lookup user by email logic working for data sharing. For this we'll need to store a token for the management API. Do you have any insight into how to persist a JWT for the api to communicate with Auth0, so we're not retrieving a new JWT for each request?

alorenzo175:
For the management API, we should be able to enable M2M communication with a new auth0 app for the sfa API. Then we can use the client id and secret to authenticate with the management API and store the token in redis (perhaps in a different db from the queue).

lboeman:
Additional updates on this PR since last review:
- On user creation, add a role with permissions to read the role and user and add it to the user. This Role exists in the Unaffiliated organization so that no other users may be granted update role permissions.
- add  a 'Read this role' permission to new roles by default.
- Allow sharing of permissions with 'read' action outside organization, so that roles can still be shared.

datastore/migrations/0023_data_sharing.up.sql
```diff
@@ -0,0 +1,466 @@
...
+READS SQL DATA SQL SECURITY DEFINER
+BEGIN
+    DECLARE orgid BINARY(16);
+    SET orgid = get_object_organization(roleid, 'roles');
+    RETURN (SELECT EXISTS(
+        SELECT 1 FROM arbiter_data.user_role_mapping
+            WHERE role_id = roleid
+           AND user_id IN (
+                SELECT id FROM arbiter_data.users
+                WHERE organization_id != orgid
```
alorenzo175: Hmm, just noticing that this has to reach basically the entire users table. Maybe the query optimizer takes care of it? maybe better to instead get all the users from user_role_mapping for role_id then query users table with user ids and org != orgid

datastore/migrations/0023_data_sharing.up.sql
```diff
@@ -0,0 +1,466 @@
...
+	DECLARE permid BINARY(16);
+    DECLARE allowed BOOLEAN DEFAULT FALSE;
+    SET allowed = user_can_create(auth0id, 'roles');
+    IF allowed THEN
+        SELECT get_user_organization(auth0id) INTO orgid;
+		SET roleid = UUID_TO_BIN(strid, 1);
+        INSERT INTO arbiter_data.roles(
+            name, description, id, organization_id) VALUES (
+            name, description, roleid, orgid);
+		SET permid = UUID_TO_BIN(UUID(), 1);
```
alorenzo175: can you fix the indentation here. just for easy reading

datastore/migrations/0023_data_sharing.up.sql
```diff
@@ -0,0 +1,466 @@
...
+BEGIN
+    DECLARE done INT DEFAULT FALSE;
+    DECLARE userid BINARY(16);
+
+    DECLARE cur CURSOR FOR SELECT id FROM arbiter_data.users;
+    DECLARE CONTINUE HANDLER FOR NOT FOUND SET done = TRUE;
+
+    OPEN cur;
+
+    read_loop: LOOP
```
alorenzo175: fun

datastore/migrations/0029_test_perms.down.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM arbiter_data.permissions WHERE description = "Role Granter";
```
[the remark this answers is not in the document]
lboeman: Just tested, it does work. I did neglect to add the drops here though. So I'll add them.

sfa_api/utils/tests/test_auth.py
```diff
@@ -66,3 +74,49 @@ def func():
...
+                mock_storage.create_new_user)
+    return fn
+
+
+@pytest.mark.parametrize('user_exists,verified,no_auth0', [
+    (True, False, False),
+    (False, False, True),
+    (False, True, False),
+    (False, False, False),
+])
```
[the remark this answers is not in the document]
lboeman: Because these parameters are actually setting the state of the mocked object above, and some of those combinations are an invalid state that the code being tested would avoid.

datastore/migrations/0023_data_sharing.up.sql
alorenzo175: Nevermind. I forgot that EXISTS stops when it finds the first row and mysql is able to efficiently rewrite the query to use the right indices

lboeman:
Okay I think this is ready. In sumarry: 
- Creates an 'Unaffiliated' organization. This is the default organization for new users, and default roles and permissions are owned by this organization. (user can read itself, and the role containing permissions to read itself).
- Allows admin users to grant roles outside their organization as long as the grantee's organization has accepted the TOU.
- Roles contain a permission to read themselves on creation.
- Reading a single role now includes a list of user ids that have been granted that role. 
- Creates users if they have a valid token and email is verified. Will have to update the dashboard for better handling. Currently if a user signs up but does not have a verified email, a json decode error is thrown in many cases.

sfa_api/utils/auth.py
```diff
@@ -46,7 +48,66 @@ def verify_access_token():
...
+    Returns
+    -------
+    unser_info: dict
+        Dict of user information provided by auth0 or None on failure
+    Raises
+    ------
+    json.decoder.JSONDecodeError
+        If the user info request fails.
+    """
+    info_request = requests.get(
```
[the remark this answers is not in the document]
alorenzo175: needs a simple test

sfa_api/utils/storage_interface.py
```diff
@@ -1354,3 +1354,21 @@ def store_report_status(report_id, status):
...
     """
     _call_procedure('store_report_status', report_id, status)
+
+
+def get_current_user_info():
+    user_info = _call_procedure_for_single('get_current_user_info')
+    return user_info
+
+
+def create_new_user():
```
alorenzo175: add a test

alorenzo175:
:+1:

PR #153, opened 2019-09-06
Fix minor user management issues introduced by 145

lboeman:


lboeman:
Fixed an issue with not serializing the user list returned with roles correctly, and handled IntegrityErrors from trying to insert the same user-role or role-permission mapping more than once.

sfa_api/utils/storage_interface.py
```diff
@@ -903,9 +904,15 @@ def add_role_to_user(user_id, role_id):
           permissions to read role and user.
         - If the calling user does not have
           permission to update the user.
+        - If the user has already been granted
```
alorenzo175: This one is a BadAPIRequest not StorageAuthError

sfa_api/utils/storage_interface.py
```diff
@@ -1000,11 +1008,15 @@ def add_permission_to_role(role_id, permission_id):
     ------
     StorageAuthError
         - If the user does not have permission to update the role.
+        - If the role already contains the permission.
         - If the role or permission does not exist.
-        - If the iser does not have permission to read the role and
+        - If the user does not have permission to read the role and
```
alorenzo175: again badapi not storageautherror

sfa_api/utils/storage_interface.py
```diff
@@ -903,9 +904,15 @@ def add_role_to_user(user_id, role_id):
...
           permission to update the user.
+        - If the user has already been granted
+          the role.
     """
-    _call_procedure('add_role_to_user',
-                    user_id, role_id)
+    try:
+        _call_procedure('add_role_to_user',
+                        user_id, role_id)
+    except pymysql.err.IntegrityError:
```
alorenzo175: what else could raise an integrity error? I'm not sure, but feels like there could be other errors that we would want to know about and not catch, so I think we check the error to make sure the role is already granted.

sfa_api/tests/rbac/test_roles.py
```diff
@@ -185,3 +192,18 @@ def test_add_perm_to_role_missing_perm(api, new_role, new_perm, remove_perms):
...
+    perms = api.get('/permissions/', BASE_URL)
+    assert perm_id in [perm['permission_id'] for perm in perms.json]
+    added_perm = api.post(f'/roles/{role_id}/permissions/{perm_id}',
+                          BASE_URL)
+    assert added_perm.status_code == 204
+    added_perm = api.post(f'/roles/{role_id}/permissions/{perm_id}',
+                          BASE_URL)
+    assert added_perm.status_code == 400
+    assert added_perm.json == {"errors": {
+        "role": ["Role already contains permission."]}}
```
alorenzo175: Just to make sure there isn't an information leak, can you add a test that creates the permission, then removes permission to update roles, then tries to add the same permission again and returns 404 and not 400

sfa_api/tests/rbac/test_users.py
```diff
@@ -126,3 +126,17 @@ def test_current_user(api):
...
+    add_role = api.post(f'/users/{user_id}/roles/{role_id}', BASE_URL)
+    assert add_role.status_code == 204
+    get_user = api.get(f'/users/{user_id}', BASE_URL)
+    user = get_user.json
+    roles_on_user = user['roles'].keys()
+    assert role_id in roles_on_user
+    add_role = api.post(f'/users/{user_id}/roles/{role_id}', BASE_URL)
+    assert add_role.status_code == 400
+    assert add_role.json == {"errors": {
+        "user": ["User already granted role."]}}
```
alorenzo175: similar to above. Add a test that grants the role, removes permission, then tries to grant same again and returns 404 instead of 400

sfa_api/utils/storage_interface.py
```diff
@@ -1000,11 +1008,15 @@ def add_permission_to_role(role_id, permission_id):
...
+        - If the role already contains the permission.
         - If the role or permission does not exist.
-        - If the iser does not have permission to read the role and
+        - If the user does not have permission to read the role and
           permission.
     """
-    _call_procedure('add_permission_to_role', role_id, permission_id)
+    try:
+        _call_procedure('add_permission_to_role', role_id, permission_id)
+    except pymysql.err.IntegrityError:
```
alorenzo175: like above, if integrityerror is raise for other things, make sure we only catch it because the role already has the permission and not some other reason

Issue #141, opened 2019-08-12
Data sharing updates


lboeman:
- ~~New permissions object type for creating user-role mapping~~. *Created grant/revoke action for roles.
- ~~Add read role permission to roles on creation.~~
- ~~Api endpoints to list roles/permissions of user.~~  Single role endpoints now include role and user ids as lists.
- Setup Auth0 management API to read user data to pair with internal user data for display. 
- CLI for admin to add users to organization. Prompted by administrator's email to add a user.

lboeman:
Closing. relevant points have their own issues.

PR #155, opened 2019-09-16
Framework admin SQL 

lboeman:
Add mysql procedures for creating organizations, establishing default roles and adding/removing users from organizations.
related: #141 , #156

### DIFF
--- a/datastore/migrations/0024_framework_admin.up.sql
+++ b/datastore/migrations/0024_framework_admin.up.sql
@@ -494,3 +494,27 @@ END;
 GRANT SELECT (id), DELETE ON arbiter_data.users TO 'delete_rbac'@'localhost';
 GRANT EXECUTE ON PROCEDURE arbiter_data.delete_user TO 'delete_rbac'@'localhost';
 GRANT EXECUTE ON PROCEDURE arbiter_data.delete_user TO 'frameworkadmin'@'%';
+
+
+/*
+ * List all users in the framework
+ */
+CREATE DEFINER = 'select_rbac'@'localhost' PROCEDURE list_all_users()
+READS SQL DATA SQL SECURITY DEFINER
+BEGIN
+    SELECT auth0_id, BIN_TO_UUID(id, 1), get_organization_name(organization_id), organization_id FROM arbiter_data.users;
+END;
+GRANT EXECUTE ON PROCEDURE arbiter_data.list_all_users TO 'select_rbac'@'localhost';
+GRANT EXECUTE ON PROCEDURE arbiter_data.list_all_users TO 'frameworkadmin'@'%';
+
+
+/*
+ * List all organizations in the framework
+ */
+CREATE DEFINER = 'select_rbac'@'localhost' PROCEDURE list_all_organizations()
+READS SQL DATA SQL SECURITY DEFINER
+BEGIN
+    SELECT name, BIN_TO_UUID(id, 1) from arbiter_data.organizations;
+END;
+GRANT EXECUTE ON PROCEDURE arbiter_data.list_all_organizations TO 'select_rbac'@'localhost';
+GRANT EXECUTE ON PROCEDURE arbiter_data.list_all_organizations TO 'frameworkadmin'@'%';
--- a/datastore/migrations/0024_framework_admin.up.sql
+++ b/datastore/migrations/0024_framework_admin.up.sql
@@ -512,7 +512,7 @@ GRANT EXECUTE ON PROCEDURE arbiter_data.list_all_users TO 'frameworkadmin'@'%';
 CREATE DEFINER = 'select_rbac'@'localhost' PROCEDURE list_all_organizations()
 READS SQL DATA SQL SECURITY DEFINER
 BEGIN
-    SELECT name, BIN_TO_UUID(id, 1) as id from arbiter_data.organizations;
+    SELECT name, BIN_TO_UUID(id, 1) as id, accepted_tou from arbiter_data.organizations;
 END;
 GRANT EXECUTE ON PROCEDURE arbiter_data.list_all_organizations TO 'select_rbac'@'localhost';
 GRANT EXECUTE ON PROCEDURE arbiter_data.list_all_organizations TO 'frameworkadmin'@'%';
--- a/datastore/migrations/0024_framework_admin.up.sql
+++ b/datastore/migrations/0024_framework_admin.up.sql
@@ -426,7 +426,7 @@ BEGIN
         MYSQL_ERRNO = 1142;
     END IF;
 END;
-GRANT SELECT, UPDATE ON arbiter_data.users TO 'update_rbac'@'localhost';
+GRANT SELECT(id), UPDATE ON arbiter_data.users TO 'update_rbac'@'localhost';
 GRANT EXECUTE ON FUNCTION arbiter_data.get_object_organization TO 'update_rbac'@'localhost';
 GRANT EXECUTE ON PROCEDURE arbiter_data.add_user_to_org TO 'update_rbac'@'localhost';
 GRANT EXECUTE ON PROCEDURE arbiter_data.add_user_to_org TO 'frameworkadmin'@'%';
@@ -484,7 +484,7 @@ MODIFIES SQL DATA SQL SECURITY DEFINER
 BEGIN
     DECLARE userid BINARY(16);
     SET userid = UUID_TO_BIN(struserid, 1);
-    IF EXISTS(SELECT * FROM arbiter_data.users WHERE id = userid) THEN
+    IF EXISTS(SELECT 1 FROM arbiter_data.users WHERE id = userid) THEN
         DELETE FROM arbiter_data.users WHERE id = userid;
     ELSE
         SIGNAL SQLSTATE '42000' SET MESSAGE_TEXT = 'User does not exist',
@@ -532,13 +532,13 @@ MODIFIES SQL DATA SQL SECURITY DEFINER
 BEGIN
     DECLARE orgid BINARY(16);
     SET orgid = UUID_TO_BIN(strorgid, 1);
-    IF EXISTS(SELECT * FROM arbiter_data.organizations WHERE id = orgid) THEN
+    IF EXISTS(SELECT 1 FROM arbiter_data.organizations WHERE id = orgid) THEN
         UPDATE arbiter_data.organizations SET accepted_tou = TRUE WHERE id = orgid;
     ELSE
         SIGNAL SQLSTATE '42000' SET MESSAGE_TEXT = 'Organization does not exist',
         MYSQL_ERRNO = 1305;
     END IF;
 END;
-GRANT SELECT, UPDATE ON arbiter_data.organizations TO 'update_rbac'@'localhost';
+GRANT SELECT (id), UPDATE ON arbiter_data.organizations TO 'update_rbac'@'localhost';
 GRANT EXECUTE ON PROCEDURE arbiter_data.set_org_accepted_tou TO 'update_rbac'@'localhost';
 GRANT EXECUTE ON PROCEDURE arbiter_data.set_org_accepted_tou TO 'frameworkadmin'@'%';
--- a/datastore/migrations/0024_framework_admin.up.sql
+++ b/datastore/migrations/0024_framework_admin.up.sql
@@ -174,6 +174,7 @@ BEGIN
     DECLARE create_cdf BINARY(16);
     DECLARE create_aggregates BINARY(16);
     DECLARE create_reports BINARY(16);
+    DECLARE create_sites BINARY(16);
     -- Create all types of metadata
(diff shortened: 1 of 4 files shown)


lboeman:
@alorenzo175 can you take a look at these procedures and let me know if you think there's any functionality I've missed?

datastore/migrations/0024_framework_admin.up.sql
```diff
@@ -0,0 +1,442 @@
...
+        revoke_roles, "Revoke roles", orgid, "revoke", "roles", TRUE);
+    -- update roles
+    SET update_roles = (SELECT UUID_TO_BIN(UUID(), 1));
+    INSERT INTO arbiter_data.permissions (id, description, organization_id, action, object_type, applies_to_all) VALUES (
+        update_roles, "Update roles", orgid, "update", "roles", TRUE);
+    -- update permissions
+    SET update_permissions = (SELECT UUID_TO_BIN(UUID(), 1));
+    INSERT INTO arbiter_data.permissions (id, description, organization_id, action, object_type, applies_to_all) VALUES (
+        update_permissions, "Update permissions", orgid, "update", "permissions", TRUE);
+    INSERT INTO arbiter_data.role_permission_mapping(role_id, permission_id) VALUES (
```
alorenzo175: no delete?

datastore/migrations/0024_framework_admin.up.sql
```diff
@@ -0,0 +1,442 @@
...
+    DECLARE orgid BINARY(16);
+    DECLARE userid BINARY(16);
+    SET orgid = UUID_TO_BIN(strorgid, 1);
+    SET userid = UUID_TO_BIN(struserid, 1);
+    IF orgid = get_object_organization(userid, 'users') THEN
+        -- add all default roles to user
+        INSERT INTO arbiter_data.user_role_mapping(user_id, role_id) VALUES(
+            userid, get_org_role_by_name('Create metadata', orgid)), (
+            userid, get_org_role_by_name('Read all', orgid)), (
+            userid, get_org_role_by_name('Write all values', orgid)), (
```
alorenzo175: some are missing, is that on purpose?

datastore/migrations/0024_framework_admin.up.sql
```diff
@@ -0,0 +1,442 @@
...
+GRANT SELECT, UPDATE ON arbiter_data.users TO 'update_rbac'@'localhost';
+GRANT EXECUTE ON FUNCTION arbiter_data.get_object_organization TO 'update_rbac'@'localhost';
+GRANT EXECUTE ON PROCEDURE arbiter_data.add_user_to_org TO 'update_rbac'@'localhost';
+GRANT EXECUTE ON PROCEDURE arbiter_data.add_user_to_org TO 'frameworkadmin'@'%';
+
+
+/*
+ * Remove all roles 
+ */
+CREATE DEFINER = 'delete_rbac'@'localhost' PROCEDURE remove_org_roles_from_user(
```
alorenzo175: this is remove all roles from outside organizations, right? maybe the name should be more clear

datastore/migrations/0024_framework_admin.up.sql
```diff
@@ -0,0 +1,442 @@
...
+CREATE DEFINER = 'insert_rbac'@'localhost' PROCEDURE remove_user_from_org(
+    IN struserid CHAR(36), IN strorgid CHAR(36))
+MODIFIES SQL DATA SQL SECURITY DEFINER
+BEGIN
+    DECLARE orgid BINARY(16);
+    DECLARE userid BINARY(16);
+    SET orgid = UUID_TO_BIN(strorgid, 1);
+    SET userid = UUID_TO_BIN(struserid, 1);
+    -- remove all non-unaffiliated Organizational roles from the user
+    CALL remove_org_roles_from_user(userid, orgid);
```
alorenzo175: this removes all non orgid roles I think. orgid here should be the unaffiliated id

datastore/migrations/0024_framework_admin.up.sql
```diff
@@ -0,0 +1,442 @@
...
+    UPDATE arbiter_data.users SET organization_id = get_unaffiliated_orgid() WHERE id = userid;
+END;
+GRANT EXECUTE ON PROCEDURE move_user_to_unaffiliated TO 'update_rbac'@'localhost';
+GRANT EXECUTE ON PROCEDURE move_user_to_unaffiliated TO 'insert_rbac'@'localhost';
+
+
+/*
+ * Remove user from organization
+ */
+CREATE DEFINER = 'insert_rbac'@'localhost' PROCEDURE remove_user_from_org(
```
alorenzo175: why insert_rbac? seems like update_rbac is more appropriate

datastore/migrations/0024_framework_admin.up.sql
```diff
@@ -0,0 +1,442 @@
...
+MODIFIES SQL DATA SQL SECURITY DEFINER
+BEGIN
+    DECLARE orgid BINARY(16);
+    DECLARE userid BINARY(16);
+    SET orgid = UUID_TO_BIN(strorgid, 1);
+    SET userid = UUID_TO_BIN(struserid, 1);
+    -- remove all non-unaffiliated Organizational roles from the user
+    CALL remove_org_roles_from_user(userid, orgid);
+    -- update user's organization to Unaffiliated
+    CALL move_user_to_unaffiliated(userid);
```
alorenzo175: I think this should just be inlined here and deleted above just remove the chance that a user is switch to unaffiliated but the roles remain

datastore/migrations/0024_framework_admin.up.sql
alorenzo175: was this previously called move_user_to_unaffiliated like in down? I like that better

datastore/migrations/0024_framework_admin.up.sql
lboeman: Ahh, good catch, I also neglected to add read permissions here.

lboeman:
Thanks for the review, I'll work on the tests.

lboeman:
Added a couple of utility functions to support the CLI, and am updating `remove_external_org_roles` to a less generic procedure that won't eliminate access to the reference org as well. Oversight on my part.

datastore/migrations/0024_framework_admin.up.sql
```diff
@@ -0,0 +1,515 @@
...
+    SET create_sites = UUID_TO_BIN(UUID(), 1);
+     INSERT INTO arbiter_data.permissions (id, description, organization_id, action, object_type, applies_to_all) VALUES (
+        create_sites, "Create sites", orgid, "create", "sites", TRUE);
+    -- add all write perms to the role
+    INSERT INTO arbiter_data.role_permission_mapping (role_id, permission_id) VALUES (
+        roleid, create_obs), (
+        roleid, create_fx), (
+        roleid, create_cdf), (
+        roleid, create_aggregates), (
+        roleid, create_reports);
```
alorenzo175: missing create_sites

datastore/migrations/0024_framework_admin.up.sql
```diff
@@ -0,0 +1,515 @@
...
+    SET delete_sites = UUID_TO_BIN(UUID(), 1);
+    INSERT INTO arbiter_data.permissions (id, description, organization_id, action, object_type, applies_to_all) VALUES (
+        delete_sites, "Delete sites", orgid, "delete", "sites", TRUE);
+    -- add all write perms to the role
+    INSERT INTO arbiter_data.role_permission_mapping (role_id, permission_id) VALUES (
+        roleid, delete_obs), (
+        roleid, delete_fx), (
+        roleid, delete_cdf), (
+        roleid, delete_aggregates), (
+        roleid, delete_reports);
```
alorenzo175: missing delete_sites

datastore/migrations/0024_framework_admin.up.sql
```diff
@@ -0,0 +1,515 @@
...
+/*
+ * Get a role id by name and organization
+ */
+CREATE DEFINER = 'select_rbac'@'localhost' FUNCTION get_org_role_by_name(
+    role_name VARCHAR(64), orgid BINARY(16))
+RETURNS BINARY(16)
+READS SQL DATA SQL SECURITY DEFINER
+BEGIN
+    RETURN (SELECT id FROM arbiter_data.roles
+        WHERE organization_id = orgid AND name = role_name LIMIT 1);
```
alorenzo175: I think we should do a bit more handling here and raise an error if there is more than one role with role_name.

datastore/tests/test_admin_procedures.py
```diff
@@ -0,0 +1,641 @@
...
+    dictcursor.execute(
+        'SELECT * FROM user_role_mapping WHERE user_id = %s',
+        (user['id'],))
+    assert len(dictcursor.fetchall()) == 8
+    dictcursor.callproc(
+        'remove_org_roles_from_user', (user['id'],))
+    dictcursor.execute(
+        'SELECT * FROM user_role_mapping WHERE user_id = %s',
+        (user['id'],))
+    assert len(dictcursor.fetchall()) == 2
```
alorenzo175: should check that these two are unaffiliated and reference

datastore/migrations/0024_framework_admin.up.sql
lboeman: I've altered the table at the top of the file to make (organization_id, name) unique, so there shouldn't ever be two roles with the same name within the same organization. Perhaps this `LIMIT 1` isn't necessary. What kind of handling did you have in mind?

alorenzo175: ah, good that will take care of it without anything else

lboeman:
> I this ready after fixing a few things or is more needed for the CLI? We'll have to modify the openshift deployment before/when this gets merged

I'll update to these comments and add a 'org_accepted_tou' procedure. Perhaps there should be a delete organization procedure as well? Other than that this will be good to go.

lboeman:
Added accepted_tou to the organization listing so that we can view it in the tables, and verify that changes are made. also sets an error state when trying to update the acceptance of an org that doesn't exist. I think this is ready and feature complete for the first version of the admin cli.

datastore/migrations/0024_framework_admin.up.sql
```diff
@@ -0,0 +1,544 @@
...
+    -- ensure the user belongs to Unaffiliated org
+    IF get_organization_id('Unaffiliated') = get_object_organization(userid, 'users') THEN
+        UPDATE arbiter_data.users SET organization_id = orgid WHERE id = userid;
+    ELSE
+        -- error user in organization
+        SIGNAL SQLSTATE '42000' SET MESSAGE_TEXT = 'Cannot add affiliated user to organization',
+        MYSQL_ERRNO = 1142;
+    END IF;
+END;
+GRANT SELECT, UPDATE ON arbiter_data.users TO 'update_rbac'@'localhost';
```
alorenzo175: only need select(id)

datastore/migrations/0024_framework_admin.up.sql
```diff
@@ -0,0 +1,544 @@
...
+ * delete user
+ * perhaps using this is better than removing a user from an org
+ */ 
+CREATE DEFINER = 'delete_rbac'@'localhost' PROCEDURE delete_user(
+    IN struserid CHAR(36))
+MODIFIES SQL DATA SQL SECURITY DEFINER
+BEGIN
+    DECLARE userid BINARY(16);
+    SET userid = UUID_TO_BIN(struserid, 1);
+    IF EXISTS(SELECT * FROM arbiter_data.users WHERE id = userid) THEN
```
alorenzo175: select 1 instead of *

datastore/migrations/0024_framework_admin.up.sql
```diff
@@ -0,0 +1,544 @@
...
+/*
+ * Updates an organization to have accepted terms of use
+ */
+CREATE DEFINER = 'update_rbac'@'localhost' PROCEDURE set_org_accepted_tou(
+    IN strorgid CHAR(36))
+MODIFIES SQL DATA SQL SECURITY DEFINER
+BEGIN
+    DECLARE orgid BINARY(16);
+    SET orgid = UUID_TO_BIN(strorgid, 1);
+    IF EXISTS(SELECT * FROM arbiter_data.organizations WHERE id = orgid) THEN
```
alorenzo175: select 1 instead of *

datastore/migrations/0024_framework_admin.up.sql
```diff
@@ -0,0 +1,544 @@
...
+        UPDATE arbiter_data.organizations SET accepted_tou = TRUE WHERE id = orgid;
+    ELSE
+        SIGNAL SQLSTATE '42000' SET MESSAGE_TEXT = 'Organization does not exist',
+        MYSQL_ERRNO = 1305;
+    END IF;
+END;
+GRANT SELECT, UPDATE ON arbiter_data.organizations TO 'update_rbac'@'localhost';
```
alorenzo175: only need select(id) after switching to select 1 above

PR #156, opened 2019-09-19
Admin cli

### LINKED ISSUE #150: Framework Admin CLI
Should allow a framework admin to: 

- [ ] Create organization
- [ ] Add user to organziation
- [ ] Remove a user from an organization (And all roles)
- [ ] Ban/delete user

lboeman:
Adds a basic admin cli using the built in flask cli so we can use the storage interface. Still somewhat rough, needs some refactoring, ideally a way to repeat command arguments/options. Will require #155, needs to be rebased after that's merged. 
Running requires installing the API and setting the `FLASK_APP` environment variable to `sfa_api.admincli` and the `MYSQL_*` variables appropriately. Then the cli can be run with `flask admin <command>`.
Testing would require setting the `SFA_CLI_CONFIG`  env var to `AdminTestConfig`, or setting the environment variables individually.
closes #150

### DIFF
--- a/sfa_api/admincli.py
+++ b/sfa_api/admincli.py
@@ -22,6 +22,9 @@ def create_organization(organization_name):
     try:
         storage._call_procedure_without_user(
             'create_organization', organization_name)
+    except pymysql.err.DataError as e:
+        click.echo(e.args[1])
+        return
     except pymysql.err.IntegrityError as e:
         if e.args[0] == 1062:
             click.echo(f'Organization {organization_name} alread exists.')
--- a/sfa_api/admincli.py
+++ b/sfa_api/admincli.py
@@ -10,14 +10,19 @@ import pymysql
 from sfa_api import create_app
 from sfa_api.utils.errors import StorageAuthError
 
+
 config = os.getenv('SFA_CLI_CONFIG', 'ProductionConfig')
 app = create_app(config)
-admin_cli = AppGroup('admin')
+admin_cli = AppGroup(
+    'admin',
+    help="Tool for administering the Solar Forecast Arbiter Framework")
 
 
 @admin_cli.command('create-organization')
 @click.argument('organization_name', required=True)
 def create_organization(organization_name):
+    """Creates a new organization.
+    """
     from sfa_api.utils.storage import get_storage
     storage = get_storage()
     try:
@@ -39,6 +44,10 @@ def create_organization(organization_name):
 @click.argument('user_id', required=True)
 @click.argument('organization_id', required=True)
 def add_user_to_org(user_id, organization_id):
+    """
+    Adds a user to an organization. The user must currently be
+    unaffiliated.
+    """
     try:
         uuid.UUID(user_id, version=1)
     except ValueError:
@@ -65,6 +74,9 @@ def add_user_to_org(user_id, organization_id):
 @click.argument('user_id', required=True)
 @click.argument('organization_id', required=True)
 def promote_to_admin(user_id, organization_id):
+    """
+    Grants a user admin permissions in the organizations.
+    """
     try:
         uuid.UUID(user_id, version=1)
     except ValueError:
@@ -81,13 +93,17 @@ def promote_to_admin(user_id, organization_id):
         storage._call_procedure_without_user(
             'promote_user_to_org_admin',
             user_id, organization_id)
-    except Exception as e:
+    except StorageAuthError as e:
         click.echo(e.args[0])
 
 
 @admin_cli.command('move-to-unaffiliated')
 @click.argument('user_id', required=True)
 def move_user_to_unaffiliated(user_id):
+    """
+    Moves a user to the Unaffiliated organization and removes access
+    to all data except for the reference data set.
+    """
     try:
         uuid.UUID(user_id, version=1)
     except ValueError:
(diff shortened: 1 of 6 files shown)


lboeman:
I think this is in a decent place. Code coverage is missing spots where I've reraised an error after catching a specific mysql errno. The only issue I have with it's current iteration is that add_user_to_org checks if the user is in the Unaffiliated organization before allowing them to be added, and as a result the error returned when a user does not exist is 'Cannot add affiliated user to organization'. 
I think this is ready for review, but perhaps after being rebased upon #155

lboeman:
Added username and password as options to this as per comments from @alorenzo175 in the openshift deployment repo. Can be passed as using `--password <password>` or the user will be prompted.

sfa_api/utils/storage_interface.py
```diff
@@ -135,6 +135,19 @@ def try_query(query_cmd):
             raise
 
 
+def _call_procedure_without_user(procedure_name, *args, cursor_type='dict'):
```
alorenzo175: just add a kwarg to ``_call_procedure`` like ``with_current_user=True`` and avoid copy pasta. And make ``new_args = (current_user, *args)`` or ``new_args = args`` so you don't have to worry about ``len(args) + 1``

sfa_api/admincli.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+config = os.getenv('SFA_CLI_CONFIG', 'ProductionConfig')
+app = create_app(config)
+admin_cli = AppGroup(
+    'admin',
+    help="Tool for administering the Solar Forecast Arbiter Framework")
+
+
+username_opt = click.option('--username', required=True)
```
alorenzo175: add ``show_envvar=True, envvar='MYSQL_USER', callback=lambda x: app.config['MYSQL_USER'] = x``. This should then not require the option if MYSQL_USER is set and avoid the need for ``update_config``

sfa_api/admincli.py
```diff
@@ -0,0 +1,255 @@
...
+password_opt = click.option('--password', prompt=True, required=True,
+                            hide_input=True)
```
alorenzo175: Similar to comment for username, but for MYSQL_PASSWORD

sfa_api/admincli.py
```diff
@@ -0,0 +1,255 @@
...
+import click
+from flask.cli import AppGroup
+import pymysql
+
+
+from sfa_api import create_app
+from sfa_api.utils.errors import StorageAuthError
+
+
+config = os.getenv('SFA_CLI_CONFIG', 'ProductionConfig')
```
alorenzo175: Is there a way this can be set as a cli option? Or maybe it doesn't need to be configurable as long as where have mysql host, port, database options?

sfa_api/admincli.py
```diff
@@ -0,0 +1,255 @@
...
+app = create_app(config)
```
alorenzo175: seems a little heavy to load everything for the app, when we really just need a generic ``app = Flask(__name__)`` with a config loaded

sfa_api/admincli.py
```diff
@@ -0,0 +1,255 @@
...
+    """Creates a new organization.
+    """
+    update_config(username, password)
+    from sfa_api.utils.storage import get_storage
+    storage = get_storage()
+    try:
+        storage._call_procedure_without_user(
+            'create_organization', organization_name)
+    except pymysql.err.DataError:
+        click.echo("Organization name must be 32 characters or fewer.")
```
alorenzo175: on errors throughout this file, should ``sys.exit(1)``so you can tell in the shell that the command failed

sfa_api/admincli.py
```diff
@@ -0,0 +1,255 @@
...
+@click.argument('user_id', required=True)
+@click.argument('organization_id', required=True)
+def add_user_to_org(user_id, organization_id, username, password):
+    """
+    Adds a user to an organization. The user must currently be
+    unaffiliated.
+    """
+    update_config(username, password)
+    try:
+        uuid.UUID(user_id, version=1)
```
alorenzo175: instead of checking here, set the type to ``click.UUID`` and if that doesn't check the format, subclass it and add the check to the convert method https://click.palletsprojects.com/en/7.x/parameters/

sfa_api/admincli.py
alorenzo175: or better make a new function
```python
def fail(message):
    click.echo(message, err=True, color=if you want)
    sys.exit(1)
```

sfa_api/admincli.py
```diff
@@ -0,0 +1,255 @@
...
+    except StorageAuthError as e:
+        click.echo(e.args[0])
+    else:
+        click.echo(f'Added user {user_id} to organization {organization_id}')
+
+
+@admin_cli.command('promote-user-to-org-admin')
+@requires_auth
+@click.argument('user_id', required=True)
+@click.argument('organization_id', required=True)
```
alorenzo175: maybe it will be tedious for us, but some of these functions are important enough that we probably want to use ``click.confirmation_option`` force us to double check values

alorenzo175:
Oh, I need to set ``FLASK_APP=sfa_api.admincli`` for this to work? I'll forget that everytime. Can we add the admin group as another command under ``sfa-api`` using https://flask.palletsprojects.com/en/1.1.x/cli/#custom-scripts?

PR #159, opened 2019-09-24
Add aggregates to API

alorenzo175:


alorenzo175:
As I'm writing the tests for the database tables, I want to get some feedback on if the aggregate fields are reasonable. Right now, aggregates will be associated with a list of observations along with:
-  id: uuid
- organization_id: uuid
- name: str
- variable: str, will enforce that this is the same as all the observations
- interval_label: right now only beginning and ending. I don't think instant makes sense for aggregates or at least for ease of computing the aggregate.
- interval_length: int
- extra_parameters
- created_at: timestamp
- modified_at: timestamp

We could support an "aggregate" with ``interval_value_type`` other than sum, but maybe that is not that useful or worth supporting.

As far as permissions go, a user with ``read_values`` permissions on an aggregate will get the timeseries of each underlying observation with some random id. This means they do not have to have ``read_values``access to the observation objects, but might be able to figure out which timeseries corresponds to which observation. Therefore, we need to make sure that the api never returns the individual series and that this is clearly documented.

We also need to account for the possibility that an observation is deleted. I don't want to restrict observation deletion if an aggregate with it exists, and we probably want to support user created aggregates of observations outside their organization. So, we also need to keep track of when an observation is removed from an aggregate, and indicate (probably as a quality flag) that an observation is missing from the data if it is deleted. I think I'll also just take the logical or of all the individual quality_flags and assign that as the aggregate quality_flag.

What do you think @wholmgren @lboeman ?

wholmgren:
> interval_label: right now only beginning and ending. I don't think instant makes sense for aggregates or at least for ease of computing the aggregate.

Ok, but why not instantaneous if `interval_length` is the same for all? I guess you'd need to ensure that the actual times are still the same (e.g. some hourly data could be labeled at hour + 1 minute). 

> We could support an "aggregate" with interval_value_type other than sum, but maybe that is not that useful or worth supporting.

Average could be useful for irradiance. For example, average irradiance for all reference data in a climate zone.

> permissions

Would it simplify things if we required that the actual time series is available? In situations in which this is a problem, the data provider could be required to compute and upload their own aggregate time series. The "your time series data is anonymously available" part is going to be hard to explain and justify.

> So, we also need to keep track of when an observation is removed from an aggregate, and indicate (probably as a quality flag) that an observation is missing from the data if it is deleted.

sounds reasonable

alorenzo175:
> Ok, but why not instantaneous if `interval_length` is the same for all? I guess you'd need to ensure that the actual times are still the same (e.g. some hourly data could be labeled at hour + 1 minute).

I was going off of the description of aggregates [here](https://github.com/SolarArbiter/solarforecastarbiter-core/issues/61#issuecomment-499933670). Basically that we should support mismatched labels and intervals.

> Average could be useful for irradiance. For example, average irradiance for all reference data in a climate zone.

ok

> Would it simplify things if we required that the actual time series is available? In situations in which this is a problem, the data provider could be required to compute and upload their own aggregate time series. The "your time series data is anonymously available" part is going to be hard to explain and justify.

yes, but having a separate time series that is only associated with an aggregate complicates things. For now, easiest is you need to be able to get the values for each underlying observation if you want the aggregate values.

wholmgren:
> I was going off of the description of aggregates here. Basically that we should support mismatched labels and intervals.

Yes for interval averages, but I think that's not feasible for instantaneous. But would be nice to allow for matched instantaneous labels/intervals. Otherwise we're requiring a resample step that at best results in a mislabeling and at worse enforces undesired downsampling.

> having a separate time series that is only associated with an aggregate complicates things. For now, easiest is you need to be able to get the values for each underlying observation if you want the aggregate values.

Ok.

alorenzo175:
> Yes for interval averages, but I think that's not feasible for instantaneous. But would be nice to allow for matched instantaneous labels/intervals. Otherwise we're requiring a resample step that at best results in a mislabeling and at worse enforces undesired downsampling.

I don't think it's reasonable to enforce matching instantaneous timestamps even if interval length is the same (like your example of obs 1coming on the hour and obs 2 coming in at  hour + 1min). It wouldn't be good from a UI perspective to fail to compute that aggregate value only at the time it is requested instead of when the aggregate object is created.

 I guess a windowed approach could work i.e. aggregate value on the hour is obs 1 from hour + obs 2 from hour + 1min. But then should the window center and width be configurable?

lboeman:
> As far as permissions go, a user with `read_values` permissions on an aggregate will get the timeseries of each underlying observation with some random id. This means they do not have to have `read_values`access to the observation objects, but might be able to figure out which timeseries corresponds to which observation. Therefore, we need to make sure that the api never returns the individual series and that this is clearly documented.

So if a user has `read_values`, the database procedure returns individual series that make up the aggregate, and the python in the api/core is responsible for actually computing the aggregate? I think that sounds good.
It seems like `read` on an aggregate should provide some metadata about the observations, or the sites of said observations. Or should the permissions remain granular, and we'll leave it to the dashboard/user to look those up?

> We also need to account for the possibility that an observation is deleted. I don't want to restrict observation deletion if an aggregate with it exists, and we probably want to support user created aggregates of observations outside their organization. So, we also need to keep track of when an observation is removed from an aggregate, and indicate (probably as a quality flag) that an observation is missing from the data if it is deleted. I think I'll also just take the logical or of all the individual quality_flags and assign that as the aggregate quality_flag.

Would this quality flag just signify that the observations list has changed since the aggregates creation? Observations could be added to the aggregate as well right?

Otherwise seems very reasonable to me.

alorenzo175:
> So if a user has `read_values`, the database procedure returns individual series that make up the aggregate, and the python in the api/core is responsible for actually computing the aggregate? I think that sounds good.

Yes. Adjusting the intervals/dealing with nans would be a real pain in SQL.

> It seems like `read` on an aggregate should provide some metadata about the observations, or the sites of said observations. Or should the permissions remain granular, and we'll leave it to the dashboard/user to look those up?

Leave it to the dashboard/user. Like how roles work now with permissions.

> Would this quality flag just signify that the observations list has changed since the aggregates creation? 

It would signify that the value at a given time is missing at least one of the observations that are supposed to go into it. So if the aggregate is of obs0 and obs1, and obs0 is deleted at 2019-09-30T12:00:00Z, the aggregate for 2019-09-30T10:00Z to 2019-09-30T14:00Z might be something like

| time | value | qflag|
|-------|--------|--------|
| 2019-09-30T10:00Z | 3.0 | missing observation |
| 2019-09-30T11:00Z | 3.0 | missing observation |
| 2019-09-30T12:00Z | 3.0 | OK |
| 2019-09-30T13:00Z | 3.0 | OK |
| 2019-09-30T14:00Z | 3.0 |OK |

> Observations could be added to the aggregate as well right?

yes

alorenzo175:
@wholmgren any more thoughts on instantaneous aggregates?

wholmgren:
Is there actually a requirement that interval average data be labeled with times that are divisible by the interval length? For example, could someone upload 5 minute data labeled with times `12:01, 12:06, etc`?

alorenzo175:
As far as I can tell right now without actually testing, yes someone could do that. The data would also not be flagged because the validation only looks at the time diff between values.

wholmgren:
If that is allowed as part of the design, then similar questions apply to interval average quantities as to instantaneous. We could also document that data for use in aggregates needs to meet certain additional conditions.

alorenzo175:
@lboeman do you want to take a look at the SQL and tests now before I add in the API components and it gets more unruly?

lboeman:
> @lboeman do you want to take a look at the SQL and tests now before I add in the API components and it gets more unruly?
> 
> "now" as in before I do more, but we this can wait as we fix other bugs.

Yeah, that sounds good, I have those other PRs to wrap up, I'll take a look after that?

sfa_api/aggregates.py
```diff
@@ -0,0 +1,333 @@
...
+        """
+        ---
+        summary: Update an aggregate.
+        description: >-
+          For now, only adding or removing observations to/from the
+          aggregate is supported. If an observation is already part
+          of an aggregate, effective_until must be set until it can
+          be added again. Any attempt to set 'effective_until'
+          will apply to all observations with the given ID in the
+          aggregate.
```
lboeman: Maybe add  something like `Only one of "effective_until" or "effective_from" should be set per request.` I found this a little confusing at first.

sfa_api/aggregates.py
```diff
@@ -0,0 +1,333 @@
...
+            if 'effective_from' in update_obs:
+                obs_id = str(update_obs['observation_id'])
+                obs = storage.read_observation(obs_id)
+                agg = storage.read_aggregate(aggregate_id)
+                errors = defaultdict(partial(defaultdict, dict))
+                for aggobs in agg['observations']:
+                    if (
+                            aggobs['observation_id'] == obs_id and
+                            aggobs['effective_until'] is None
+                    ):
```
lboeman: I think this also needs an or clause with a check for `aggobs['effective_until'] > update_obs['effective_from']`. It looks like if someone were to preemptively add an `effective_until` to an observation, there would be periods where multiple copies of the same observation would be added.  Maybe that check is better implemented in SQL though?

alorenzo175: I don't think we actually want to restrict overlap of effective_until and effective_from. Say for example that I accidentally add an observation to an aggregate with the wrong ``effective_from``. I want to be able to "remove" it (set effective until) and then set ``effective_from`` on a new item in the list to before that previous effective until. I added a test to make sure this works as expected.

alorenzo175:
@lboeman I think this is finally fully ready. I'll deal with making forecasts for an aggregate in a separate PR.

lboeman:
Looks good to me! I see I misunderstood how that mysql join works for the overlapping.

Issue #162, opened 2019-09-27
Users moved to organization are unreadable by admin.


lboeman:
Since the trigger to update "applies_to_all" permissions fires when the object is inserted into the table. Moving a user from one organization to another doesn't add them to the 'read all users' permissions of the new organization.

lboeman:
@alorenzo175 I was thinking this should be handled by an trigger on updates to the user table. Does that sound reasonable to you?

alorenzo175:
yes

lboeman:
closed by #163

PR #163, opened 2019-09-28
Add trigger on update users to remove permissions from previous orgs

lboeman:
Added to address #162 
Not sure if this should go somewhere else or if it can be tacked on to the framework admin sql migration.

### DIFF
--- a/datastore/migrations/0024_framework_admin.up.sql
+++ b/datastore/migrations/0024_framework_admin.up.sql
@@ -573,6 +573,7 @@ BEGIN
             SELECT id, NEW.id FROM arbiter_data.permissions
             WHERE organization_id = NEW.organization_id
                 AND object_type = 'users'
+                AND action != 'create'
                 AND applies_to_all;
         CALL create_default_user_role(NEW.id, NEW.organization_id);
     END IF;
--- a/datastore/migrations/0024_framework_admin.up.sql
+++ b/datastore/migrations/0024_framework_admin.up.sql
@@ -549,11 +549,11 @@ CREATE DEFINER = 'delete_rbac'@'localhost' PROCEDURE remove_user_facing_permissi
 SQL SECURITY DEFINER MODIFIES SQL DATA
 BEGIN
     -- Remove all mappings that grant actions on the user
-    DELETE FROM arbiter_data.permission_object_mapping WHERE object_id = NEW.id;
+    DELETE FROM arbiter_data.permission_object_mapping WHERE object_id = userid;
     -- delete the default user role and specific default permissions
-    DELETE FROM arbiter_data.roles WHERE name = CONCAT('User role', BIN_TO_UUID(NEW.id, 1));
-    DELETE FROM arbiter_data.permissions WHERE description = CONCAT('Read Self User ', BIN_TO_UUID(NEW.id, 1));
-    DELETE FROM arbiter_data.permissions WHERE description = CONCAT('Read User Role ', BIN_TO_UUID(NEW.id, 1));
+    DELETE FROM arbiter_data.roles WHERE name = CONCAT('User role', BIN_TO_UUID(userid, 1));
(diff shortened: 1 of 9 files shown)


datastore/migrations/0024_framework_admin.up.sql
```diff
@@ -542,3 +542,28 @@ END;
...
+            WHERE object_id = NEW.id AND permission_id IN (
+                SELECT id FROM arbiter_data.permissions
+                WHERE organization_id = OLD.organization_id
+                    AND object_type = 'users'
+                    AND action in ('read', 'update'));
+        END IF;
+        INSERT INTO arbiter_data.permission_object_mapping (permission_id, object_id)
+            SELECT id, NEW.id FROM arbiter_data.permissions
+            WHERE organization_id = NEW.organization_id
+                AND object_type = 'users'
```
alorenzo175: and ``action != 'create'``

datastore/migrations/0024_framework_admin.up.sql
```diff
@@ -542,3 +542,28 @@ END;
...
+
+
+/*
+ * Add the user to be readable by org admin after joining org
+ */
+CREATE DEFINER = 'permission_trig'@'localhost' TRIGGER update_user_perm_on_org_change AFTER UPDATE ON arbiter_data.users
+FOR EACH ROW
+BEGIN
+    IF NEW.organization_id != OLD.organization_id THEN
+        if OLD.organization_id != get_organization_id('Unaffiliated') THEN
```
alorenzo175: hmmm, excepting unaffiliated is to keep the default user self role/perms? Looking at that more carefully, not sure I agree that the roles and permissions should stay with unaffiliated instead of being recreated in the new org

lboeman:
I updated the `create_default_user_role` procedure to take an organization id as an argument and create a role there rather than unaffiliated. I had to touch 23 to do so, but I don't think there's been a tagged version since it was introduced so I think that's okay. 

Added a procedure to remove default user roles and permissions before recreation in the new org.

I've omitted an organization_id check in when deleting default roles to avoid an error condition  where an Admin user in the new organization has created a role/permission with the same name as our default roles, which would cause `add_user_to_org` to fail.

@alorenzo175 before I add tests to this would you mind taking a quick look to see how you feel about this approach?

datastore/migrations/0024_framework_admin.up.sql
```diff
@@ -542,3 +542,38 @@ END;
...
+
+
+CREATE DEFINER = 'delete_rbac'@'localhost' PROCEDURE remove_user_facing_permissions_and_default_roles(
+    IN userid BINARY(16))
+SQL SECURITY DEFINER MODIFIES SQL DATA
+BEGIN
+    -- Remove all mappings that grant actions on the user
+    DELETE FROM arbiter_data.permission_object_mapping WHERE object_id = NEW.id;
+    -- delete the default user role and specific default permissions
+    DELETE FROM arbiter_data.roles WHERE name = CONCAT('User role', BIN_TO_UUID(NEW.id, 1));
```
alorenzo175: add a space:  'User role' => 'User role '

lboeman:
I think a DEFAULT prefix would be helpful for the general understanding of the user. I'll add that in there. :+1:

datastore/tests/test_permission_triggers.py
```diff
@@ -146,3 +148,50 @@ def test_object_mapping_remove_obj(cursor, action, getfcn, new_permission,
...
     assert cursor.fetchone() is None
+
+
+# Note: This test depends on the create_default_user_role
+# function found in 0024_framework_admin.up.sql
+def test_update_user_perm_on_org_change(
+        dictcursor, new_organization, new_user):
+    user = new_user()
+    org = new_organization()
+    dictcursor.callproc(
```
alorenzo175: add ``assert user['organization_id'] != org['id']`` just to make sure they were different from the start

datastore/migrations/0024_framework_admin.up.sql
alorenzo175: still need this

PR #169, opened 2019-10-03
Validate data uploads have the same interval as interval_length

### LINKED ISSUE #45: enforce regular data uploads
once the data model is redefined for observations

alorenzo175:
closes #45

The offset issue could still be a problem e.g. I can upload data from 12:05 to 12:15 with 5 minute intervals just fine, but I can then upload data with timestamps 12:16, 12:21, 12:26, which will likely cause downstream problems. I can also always upload a single timestamp like 12:17:42.


wholmgren:
What if we also compare the last time in the database to the first time in the upload?

alorenzo175:
would need something like /latest https://github.com/SolarArbiter/solarforecastarbiter-api/issues/168

and then data uploads would require 'write_values', 'read' and 'read_values' permissions on the object. Perhaps OK, perhaps not. I guess we could get creative and add a 'read_last_value' permission or something to restrict that a bit more.

wholmgren:
The permission complication is gross. Would be better to add a new metadata
field if necessary.

wholmgren:
Ok

alorenzo175:
I think I'll allow the new timestamps to be some multiple of interval_length from the last timestamp. It's easier for users at the cost of maintaining a continuous timeseries in the database.

alorenzo175:
Too much of a pain for now. Some tests basically rely on data posted to both the demo and real api

Issue #165, opened 2019-10-02
raise errors in demo storage interface


alorenzo175:
and make it consistent with the actual SQL interface. For example, the demo interface does not raise a ``StorageAuthError`` when there is an invalid ``site_id`` when posting an observation. This requires code https://github.com/SolarArbiter/solarforecastarbiter-api/blob/7e5340526132e2d3575583f8dc93af775adf00d5/sfa_api/observations.py#L82 that is only executed for the demo interface. This is confusing.

lboeman:
Is it still valuable to maintain a demo API? I think we can probably scrap the demo api endpoints and keep the data set for testing against. There are already missing endpoints in the demo api, and it's really just a hold over from before the database was ready to be used.

PR #176, opened 2019-10-25
Allow forecasts of aggregates

alorenzo175:
So that a forecast must have either site_id or aggregate_id. Also removes the demo storage interface closing #165 

Wait to rebase on #159


wholmgren:
😲 the line diff metrics are really bad for quarterly reports

alorenzo175:
It just means there are more lines of tests relative to lines of code

PR #182, opened 2019-11-07
Add update permission to user during CDF group/report creation

### LINKED ISSUE #171: The default create role needs update on reports
Otherwise they cannot be generated and the permission  can't currently be added in the dashboard?

alorenzo175:
closes #171

### DIFF
--- a/datastore/migrations/0028_update_after_create.down.sql
+++ b/datastore/migrations/0028_update_after_create.down.sql
@@ -1,0 +1,47 @@
+DROP FUNCTION get_default_user_role;
+DROP PROCEDURE add_object_permission_to_default_user_role;
+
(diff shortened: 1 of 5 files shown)


alorenzo175:
Currently too difficult to test storage_interface/the api endpoints for the change since the user already has update permissions on everything. The mysql tests should be sufficient though.

alorenzo175:
Do you mean to avoid roles with the same name? Right now the role name is UNIQUE so mysql won't let that happen.

lboeman:
Yeah, I was thinking of the potential to create some role with a matching name, and have it automatically updated with permissions.
Taking another look at the workflow we have, it should never update a user that is not actively assigned a `Default user role <user_id>` and thus, in the organization. So I think my concerns are already adequately addressed :+1:

PR #183, opened 2019-11-07
Enable restrictions of forecast times

alorenzo175:
when ``"restrict_upload": true`` is in extra_parameters for a forecast (or cdf fx), only values that start with what should be the next init_time + lead_time_to_start are allowed to be posted.

I.e for a forecast with issue_time_of_day = 05:00 , run length of 6 hours, and lead_time_to_start of 1 hour, at 10:00 only a forecast series that begins with 12:00 will be allowed. And from 11:00:00.00001 to 16:59:59.99999, only a series with first time of 18:00 is allowed.

### DIFF
--- a/datastore/migrations/0025_write_metadata.up.sql
+++ b/datastore/migrations/0025_write_metadata.up.sql
@@ -8,6 +8,7 @@ BEGIN
     DECLARE groupid BINARY(16);
     DECLARE il INT;
     DECLARE previous_time TIMESTAMP;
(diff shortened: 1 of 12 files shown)


alorenzo175:
@wholmgren what do you think about these restrictions for trials?

wholmgren:
Makes sense. Accounts for interval label?

alorenzo175:
not yet, but it will

PR #186, opened 2019-11-08
Interface with Auth0 Mangement API

### LINKED ISSUE #142: Auth0 Management API for user information
Create a connector to the [Auth0 management API](https://auth0.com/docs/api/management/v2) to allow for searching for users by email, and displaying user data in list views.

alorenzo175:
Closes #142. Will eventually also be useful for trial things and making special framework users to run jobs.


alorenzo175:
@lboeman which endpoints need to know the user's email?

lboeman:
I think it would be good to have it on all of the `/users` get endpoints. We should add/update an endpoint to allow sharing a role with a user via their email instead of uuid.

PR #188, opened 2019-11-12
add user id -> email endpoint

lboeman:
Creates an uuid to email conversion endpoint. Requires that both the caller and target belong to organizations that have accepted the TOU. I've added the mysql to the read_user_id migration to save space, but if it should be in it's own migration I can move it.

### DIFF
--- a/datastore/migrations/0029_get_user_id.down.sql
+++ b/datastore/migrations/0029_get_user_id.down.sql
@@ -1,2 +1,2 @@
-DROP PROCEDURE IF EXISTS read_user_id;
-DROP PROCEDURE IF EXISTS read_auth0id;
+DROP PROCEDURE read_user_id;
+DROP PROCEDURE read_auth0id;
--- a/datastore/migrations/0029_get_user_id.down.sql
+++ b/datastore/migrations/0029_get_user_id.down.sql
@@ -1,1 +1,2 @@
-DROP PROCEDURE read_user_id;
+DROP PROCEDURE IF EXISTS read_user_id;
+DROP PROCEDURE IF EXISTS read_auth0id;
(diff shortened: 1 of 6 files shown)


sfa_api/utils/storage_interface.py
```diff
@@ -1756,3 +1756,25 @@ def read_user_id(auth0_id):
     """
     return _call_procedure_for_single('read_user_id', auth0_id,
                                       cursor_type='standard')[0]
+
+
+def read_auth0id(user_id):
+    """Read the auth0 id of another user. Only allowed if both users
```
alorenzo175:
```suggestion
    """Read the auth0 id of another user. Only allowed if both users are affiliated
        with organizations that have accepted the TOU.
```

datastore/migrations/0029_get_user_id.up.sql
```diff
@@ -19,3 +19,24 @@ END;
 
 GRANT EXECUTE ON PROCEDURE read_user_id TO 'select_rbac'@'localhost';
 GRANT EXECUTE ON PROCEDURE read_user_id TO 'apiuser'@'%';
+
+CREATE DEFINER = 'select_rbac'@'localhost' PROCEDURE read_auth0id (
+    IN auth0id VARCHAR(32), IN struserid CHAR(36))
+COMMENT 'Read the auth0id of another user if both orgs have accepted tou'
```
alorenzo175: need line 
READS SQL DATA SQL SECURITY DEFINER

datastore/migrations/0029_get_user_id.down.sql
```diff
@@ -1 +1,2 @@
-DROP PROCEDURE read_user_id;
+DROP PROCEDURE IF EXISTS read_user_id;
```
alorenzo175: no "if exists". Assume it does or we have some migration issue

PR #191, opened 2019-11-20
update report object pair schema

lboeman:
Started working on updating the format for object pairs. Going to need further updates, so skipping CI.


alorenzo175:
I think these test failures are from updates to core. #192 should take fix those issues

PR #194, opened 2019-11-25
Add reference jobs to migrations

alorenzo175:
For DOE RTC sites, including creating the sites/fx/obs/prob. fx, reports, and jobs for daily validation, reference nwp forecast generation, and periodic report generation. This should act as a continuous test of the job framework. This only applies to dev as we won't run this migration for prod. I'll add separate reports for prod, probably to the core repo.

Also change the 'Create reports' role to 'Recompute reports' and update permissions accordingly.


alorenzo175:
Yes, it also makes a job to generate reference forecasts for the RTC sites.

PR #195, opened 2019-11-26
Accept metrics and categories defined in core

lboeman:
Imports lists of metrics and categories from core to validate against.

### DIFF
--- a/sfa_api/schema.py
+++ b/sfa_api/schema.py
@@ -7,8 +7,7 @@ from sfa_api import spec, ma
 from sfa_api.utils.validators import (
     TimeFormat, UserstringValidator, TimezoneValidator, TimeLimitValidator)
 from solarforecastarbiter.datamodel import (
-    ALLOWED_VARIABLES, ALLOWED_CATEGORIES)
-from solarforecastarbiter.metrics.deterministic import _MAP
+    ALLOWED_VARIABLES, ALLOWED_CATEGORIES, ALLOWED_DETERMINISTIC_METRICS)
 
 
 class ISODateTime(ma.AwareDateTime):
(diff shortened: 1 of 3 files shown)


sfa_api/schema.py
```diff
@@ -6,7 +6,9 @@
 from sfa_api import spec, ma
 from sfa_api.utils.validators import (
     TimeFormat, UserstringValidator, TimezoneValidator, TimeLimitValidator)
-from solarforecastarbiter.datamodel import ALLOWED_VARIABLES
+from solarforecastarbiter.datamodel import (
+    ALLOWED_VARIABLES, ALLOWED_CATEGORIES)
+from solarforecastarbiter.metrics.deterministic import _MAP
```
alorenzo175: Can we add ALLOWED_METRICS to datamodel instead of using the "private" _MAP first?

lboeman: Yeah. The `__check_metrics__` function defers validating to the metrics package, due to the differences between probabilistic and deterministic forecasts. Not sure that it's particularly useful to the data model itself, but will be here and on the dashboard. I'll add `ALLOWED_DETERMINISTIC_METRICS` and `ALLOWED_PROBABILISTIC_METRICS`.

lboeman:
Github hasn't updated the checks section, but tests are now passing. @alorenzo175 good with you after the core update?

sfa_api/schema.py
```diff
@@ -704,12 +706,19 @@ class ReportParameters(ma.Schema):
...
     metrics = ma.List(
         ma.String(
-            validate=validate.OneOf(ALLOWED_METRICS)
+            validate=validate.OneOf(ALLOWED_REPORT_METRICS)
         ),
         title='Metrics',
         description=('The metrics to include in the report.'),
         required=True
     )
+    categories = ma.List(
```
alorenzo175: Defined twice: here and line 702

lboeman: ahh yep, thanks I botched this rebasing.

Issue #152, opened 2019-09-05
Improve failure tolerance of request_user_info


lboeman:
Used when creating a new user, to check that a user exists in auth0 and that their email is verified. This should be resilient to auth0 api failures.

lboeman:
closed by #201

PR #201, opened 2019-12-12
User info retry

lboeman:
aims at closing #152 
This is admittedly a naive approach to retrying the auth0 /userinfo endpoint., but considering this code should only run when a new user logs in for the first time, I think it should be sufficient. Also added logic such that if a 401 is returned, it returns immediately. I'm not sure if we want to add a back off or anything.


alorenzo175:
urllib3.Retry may be a more elegant way and includes backing off
See https://github.com/SolarArbiter/solarforecastarbiter-core/blob/2c422ec8aef751eff2fa86f346754f5ed2d104d1/solarforecastarbiter/io/api.py#L68-L75

PR #212, opened 2020-02-10
Speed up value inserts

alorenzo175:
It can take seconds (at least locally) to write a few thousand datapoints using one of the store_*_values procedures. This is also probably leading to a ~10 reads for every row written. Now all the values are bundled into a json array instead of going through each individual value with at least a 10x to 100x speedup.

Screenshot from 2020-02-09 21-23-25

### DIFF
--- a/sfa_api/utils/storage_interface.py
+++ b/sfa_api/utils/storage_interface.py
@@ -211,7 +211,13 @@ def _set_forecast_parameters(forecast_dict):
 def _process_df_into_json(df, rounding=8):
     """Processes a Dataframe with DatetimeIndex and 'value' column
     (with optional 'quality_flag' column) into a json string of the
-    form [{"ts": ..., "v": ...}, ...] for sending to MySQL
+    form [{"ts": ..., "v": ...}, ...] for sending to MySQL.
+
+    This numpy string processing to JSON is much faster than
+    using the json module and converting the dataframe to a list
+    of dicts. Also can't use the builtin pandas json exporter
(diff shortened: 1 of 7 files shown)


sfa_api/utils/storage_interface.py
```diff
@@ -207,6 +208,31 @@ def _set_forecast_parameters(forecast_dict):
...
+    if 'quality_flag' in df.columns:
+        qf = np.char.add(
+            ',"qf":',
+            df['quality_flag'].values.astype(int).astype(str))
+        valstr = np.char.add(valstr, qf)
+    objarr = np.char.add(timestr, valstr)
+    if len(objarr) == 0:
+        return '[]'
+    else:
+        return '[' + '},'.join(objarr) + '}]'
```
lboeman: Is this just handling types manually because the JSON doesn't hit the SQL converters?

sfa_api/utils/storage_interface.py
```diff
@@ -207,6 +208,31 @@ def _set_forecast_parameters(forecast_dict):
     return out
 
 
+def _process_df_into_json(df, rounding=8):
```
wholmgren: we should document any rounding in the dashboard and API documentation.

sfa_api/utils/storage_interface.py
alorenzo175: does the docstring change make it more clear?

lboeman: Yes, makes sense now. Thanks.

PR #216, opened 2020-02-11
Validate path UUID

### LINKED ISSUE #215: Validate UUID in path params to avoid too long strid error
Sentry Issue: [API-2V](https://sentry.io/organizations/solararbiter-rx/issues/1509203202/?referrer=github_integration)

```
DataError: (1406, "Data too long for column 'strid' at row 1")
(16 additional frame(s) were not displayed)
...
  File "pymysql/connections.py", line 732, in _read_query_result
    result.read()
  File "pymysql/connections.py", line 1075, in read
    first_packet = self.connection._read_packet()
  File "pymysql/connections.py", line 684, in _read_packet
    packet.check_error()
  File "pymysql/protocol.py", line 220, in check_error
    err.raise_mysql_exception(self._data)
  File "pymysql/err.py", line 109, in raise_mysql_exception
    raise errorclass(errno, errval)
```

lboeman:
closes #215 
Created a custom url converter because the default uuid converter actually casts to a UUID object, here we only validate on the way in. Added tests to ensure that any path added to the api with the substring '_id' returns a 404 on an invalid uuid, instead of hitting the database and possibly throwing a 500.

### DIFF
--- a/sfa_api/tests/test_api.py
+++ b/sfa_api/tests/test_api.py
@@ -1,12 +1,13 @@
 """Tests that API maintains proper state after interactions
 """
 import pytest
+import re
+import uuid
 
 
 from sfa_api.conftest import (VALID_SITE_JSON, VALID_OBS_JSON,
                               VALID_FORECAST_JSON, VALID_CDF_FORECAST_JSON,
                               BASE_URL)
-import re
 
 
 invalid_json = {'invalid': 'garbage'}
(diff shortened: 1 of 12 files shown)


sfa_api/utils/url_converters.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+from flask import abort
+from werkzeug.routing import BaseConverter
+
+
+class UUIDStringConverter(BaseConverter):
+    def to_python(self, value):
+        try:
+            return str(uuid.UUID(value, version=1))
```
alorenzo175: some testing uuids are actually UUID4, so this is modifying the string. OK (and best) to not restrict the version

sfa_api/tests/test_api.py
```diff
@@ -476,3 +477,24 @@ def test_sequence(sql_api, auth_header):
...
+    paths_with_id = []
+    for rule in sql_api.application.url_map.iter_rules():
+        if 'GET' in rule.methods and '_id' in rule.rule:
+            paths_with_id.append(re.sub('<.+>', '{id}', rule.rule))
+    return paths_with_id
+
+
+@pytest.mark.parametrize('bad_id', [
+    '56061aa6-4cf4-11ea-a21d-0a580a800331%',
+    'not a real uuid',
```
alorenzo175: add other uuid versions

lboeman: I removed the restriction on uuid version, are there still relevant uuid version tests I should check? It seems that all versions should be represented with 36 characters.

alorenzo175: nevermind this is for bad ids. I was thinking of a test of good ids which is already included in most other tests, but probably a good idea to have a separate test of the validator here with a dummy endpoint

Issue #151, opened 2019-08-30
User permissions on object endpoint


lboeman:
For creating a more user friendly dashboard. Endpoint should return a list of actions a user can perform on an object.

PR #221, opened 2020-03-05
Actions on object

lboeman:
Aimed at closing #151 . Doesn't have any testing yet, as I wanted to get some feedback on the design before investing the time. @alorenzo175 Can you take a look at this new endpoint? It lists the actions a user is allowed to take on an object. I'm foregoing the "allowed" check in sql and returning a 404 if the actions list is empty because it seems like you should be able to see your permitted actions if you have any( as opposed to say, having 'read_values' and 'write_values' and being disallowed because you don't have 'read'). This should make it such that someone can't spam requests to discover the existence of objects.

### DIFF
--- a/datastore/migrations/0036_actions_on_object.down.sql
+++ b/datastore/migrations/0036_actions_on_object.down.sql
@@ -1,0 +1,1 @@
+DROP PROCEDURE get_user_actions_on_object;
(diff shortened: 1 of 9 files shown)


alorenzo175:
I think the approach you've taken is good

PR #222, opened 2020-03-25
Recompute reports

### LINKED ISSUE #220: Add endpoint to recompute report
from the current report parameters retaining the same ID. The periodic report job does this, but would be nice to be able to do it on the fly perhaps via a button on the report page on the dashboard

lboeman:
Closes #220 
Adds a GET endpoint at `/reports/{report_id}/recompute` that queues a compute report job. 
This utilizes the `get_user_actions_on_object` procedure to check for update permission on the report before queuing the compute, but lack of other permissions could still cause the recompute to fail.


lboeman:
> good after fixing tests. Looks like an update somewhere changed message formats

I'm trying to track this down, but tests actually pass locally. I'm not sure what the issue is there.

alorenzo175:
try a ``pip install -U -r requirements.txt`` and then test

lboeman:
Was unable to actually replicate locally, but tests are passing in ci now.

PR #224, opened 2020-03-27
add /values/latest and /values/timerange endpoints

### LINKED ISSUE #168: Query for available data
We need some way of determining the date range for which data exists at a given forecast or observation. We can either implement this as a separate endpoint or include this in metadata. Could be a simple min/max query on timestamps in the values tables. 

Currently a data request returns all available data, with optional start/end parameters. Perhaps a /latest endpoint could be useful as well.

alorenzo175:
to observations, forecasts, and cdf singles. CDF groups and aggregate observations are not supported. Not sure if we should add these to `_links`

closes #168


lboeman:
>  Not sure if we should add these to `_links`

I don't think it's totally necessary since it will be in the documentation, but having the time range in links might save someone a bunch of time at some point.

PR #225, opened 2020-03-31
Update report object pair schema

lboeman:
Adds reference_forecast and uncertainty to the report object pair schema to support forecast skill and deadband uncertainty.


lboeman:
> cool, I also don't think this will harm anything in core/ loading data into the datamodel.ForecastObservation

Ahh yeah, I see now that raise on extra is not used around these. great!

PR #233, opened 2020-04-22
validate matching interval_label and variable for events

### LINKED ISSUE #232: Vaidate that event interval_label and variable match
For Forecast and Observation Schema.

lboeman:
closes #232 
Adds a validation function for calling in a `@validates_schema` method to check for the value 'event' in variable or interval label. Throws a validation error if one is found and the values aren't equal.

### DIFF
--- a/sfa_api/schema.py
+++ b/sfa_api/schema.py
@@ -503,7 +503,7 @@ class ForecastPostSchema(ma.Schema):
     extra_parameters = EXTRA_PARAMETERS_FIELD
 
     @validates_schema
-    def validate_id(self, data, **kwargs):
+    def validate_forecast(self, data, **kwargs):
         if (
                 data.get('site_id') is not None and
                 data.get('aggregate_id') is not None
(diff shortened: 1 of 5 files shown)


sfa_api/schema.py
```diff
@@ -513,6 +517,7 @@ def validate_id(self, data, **kwargs):
         ):
             raise ValidationError(
                 "One of site_id or aggregate_id must be provided")
+        validate_if_event(self, data, **kwargs)
```
alorenzo175, line 520: maybe change ``validate_id`` to ``validate_forecast``

sfa_api/tests/test_forecast.py
```diff
@@ -418,3 +418,21 @@ def test_get_forecast_timerange_404_obsid(api, observation_id):
...
+EVENT_VARIABLE = copy_update(VALID_FORECAST_JSON, 'variable', 'event')
+
+
+@pytest.mark.parametrize('payload,message', [
+    (EVENT_VARIABLE, (f'{{"events":["Both interval_label and variable must be '
+                      'set to \'event\'."]}')),
+    (EVENT_LABEL, (f'{{"events":["Both interval_label and variable must be '
+                   'set to \'event\'."]}')),
+])
+def test_observation_post_bad_event(api, payload, message):
```
alorenzo175, line 433:
```suggestion
def test_forecast_post_bad_event(api, payload, message):
```

PR #238, opened 2020-05-04
Convert temperature coefficient to %/C

alorenzo175:


sfa_api/tests/test_admincli.py
```diff
@@ -283,8 +283,8 @@ def test_set_org_accepted_tou_bad_orgid(
     result = app_cli_runner.invoke(
         admincli.set_org_accepted_tou,
         ['baduuid'] + auth_args)
-    assert ("Error: Invalid value for 'ORGANIZATION_ID': baduuid "
-            "is not a valid UUID value") in result.output
+    assert ('Error: Invalid value for "ORGANIZATION_ID": baduuid '
+            'is not a valid UUID value') in result.output
```
lboeman, line 287: I'm not entirely sure what is causing the issue with these tests. I updated with the inner double quotes to get it to pass in azure, and it works from a fresh install of the api in a new conda environment, but not by updating an existing one.

PR #243, opened 2020-05-21
Adjust job timeouts

alorenzo175:
Add the ability to adjust timeouts for the api worker, report jobs, and data validation jobs.  Also sets max content length and does not log requests by the kube probe that checks if the container is ready. 

This also adds the JOB_BASE_URL config variable that can direct jobs to use another api vs the one where the job was enqueued. The idea is that the worker pod could have an API running as a sidecar container, and that API instance can be used instead of the main instances. Then we can adjust the timeouts for that sidecar API to be long and allow for job completion even with large posts/gets. 

I also though about not using the APISession at all in the jobs like compute_report. But, the API provides a well defined interface, even if it does introduce more overhead vs direct database access, and the api and core code remain a bit more separated.

### DIFF
--- a/sfa_api/cli.py
+++ b/sfa_api/cli.py
@@ -17,6 +17,14 @@ def cli():
 
 verbose_opt = click.option('-v', '--verbose', count=True,
                            help='Increase logging verbosity')
+worker_ttl = click.option(
+    '--worker-ttl', default=420,
+    help='The timeout for the worker monitoring process'
+)
+job_monitoring_interval = click.option(
+    '--job-monitoring-interval', default=30,
+    help='The timeout for an individual job running in a work-horse process'
+)
 
 
 def _get_log_level(config, verbose, key='LOG_LEVEL'):  # pragma: no cover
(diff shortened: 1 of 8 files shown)


sfa_api/cli.py
```diff
@@ -17,6 +17,14 @@ def cli():
...
 verbose_opt = click.option('-v', '--verbose', count=True,
                            help='Increase logging verbosity')
+worker_ttl = click.option(
+    '--worker-ttl', default=420,
+    help='The timeout for the worker monitoring process'
+)
+job_monitoring_interval = click.option(
+    '--job-monitoring-interval', default=180,
+    help='The timeout for an individual job running in a work-horse process'
+)
```
lboeman, line 27: Are the help text for these backwards? I can't find actual documentation on these keyword arguments to `Worker`, but it seems like `job-monitor-interval` is the interval on which the worker receives a ping from the monitor and `default_worker_ttl` is the timeout of the job?

alorenzo175: Yeah, not sure it is in the documentation, but based on the code and experience running scheduled jobs, the descriptions are right. The worker should be able to heartbeat even when computing, but in practice it can get hung up on heavy computations which is why the default worker-ttl is high.

https://github.com/rq/rq/blob/23d43874b675a1cfc06684e9298b5079c488459c/rq/worker.py#L686
https://github.com/rq/rq/blob/23d43874b675a1cfc06684e9298b5079c488459c/rq/worker.py#L606

PR #247, opened 2020-06-01
Limit data requests

### LINKED ISSUE #239: Requesting too much data results in gateway timeout
We should place a limit on the amount of data that can be retrieved with a single request
Sentry Issue: [DASHBOARD-1X](https://sentry.io/organizations/solararbiter-rx/issues/1646612211/?referrer=github_integration)

```
HTTPError: 504 Server Error: Gateway Time-out for url: https://api.solarforecastarbiter.org/observations/063d42da-99fe-11e9-8674-0a580a8200c9/values?start=2017-05-01T17%3A04%3A00%2B00%3A00&end=2020-05-04T17%3A04%3A00%2B00%3A00
(9 additional frame(s) were not displayed)
...
  File "sfa_dash/blueprints/base.py", line 25, in insert_plot
    uuid, params={'start': start, 'end': end})
  File "sfa_dash/api_interface/observations.py", line 10, in get_values
    req = get_request(f'/observations/{observation_id}/values', **kwargs)
  File "sfa_dash/api_interface/__init__.py", line 26, in get_request
    f'{app.config["SFA_API_URL"]}{path}', **kwargs))
  File "sfa_dash/api_interface/util.py", line 62, in handle_response
    request_object.raise_for_status()
  File "requests/models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
```

alorenzo175:
Limits data requests to a configurable number of days (currently 366) and in doing so, requires the start and end query parameters. I think these are always set in core, and looks like the dashboard is ok too.

closes #239

I also created a new schema field to serialized dataframes to json via pandas directly (although it requires another load and dump by flask/marshmallow). This is much faster than dumping the frame to a dict and then dumping to json. The new way is roughly the same speed as dumping to csv.

So the things that change for users, is that start and end are required query parameters and the timestamp field will now end in 'Z' instead of '+00:00'  which is still consistent with iso8601.

### DIFF
--- a/datastore/migrations/0051_reference_jobs.up.sql
+++ b/datastore/migrations/0051_reference_jobs.up.sql
@@ -15,27 +15,27 @@ INSERT INTO sites (id, organization_id, name, latitude, longitude,
     backtrack, max_rotation_angle, dc_loss_factor, ac_loss_factor )
 VALUES (
     @abqsite, @reforgid, 'DOE RTC Albuquerque NM' , 35.050000 ,
-    -106.530000 , 1657.00 , 'America/Denver' , '{"network": "DOE RTC",
+    -106.530000 , 1657.00 , 'Etc/GMT+7' , '{"network": "DOE RTC",
     "network_api_id": "Albuquerque", "network_api_abbreviation": "",
     "observation_interval_length": 1, "attribution": "", "module":
     "Suniva 270W"}' , 0.003240 , 0.003240 , -0.420 , 'fixed' , 35.00 ,
     180.00 , NULL , NULL , NULL , NULL , NULL , 0.00 , 0.00
 ),(
     @nvsite, @reforgid, 'DOE RTC Henderson NV' , 36.040000 , -114.920000 , 538.00 ,
-    'America/Los_Angeles' , '{"network": "DOE RTC", "network_api_id":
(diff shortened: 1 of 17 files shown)


wholmgren:
Do we still need @lboeman's https://github.com/SolarArbiter/solarforecastarbiter-dashboard/pull/270 ?

Also can you roughly quantify timing for posterity and reports?

alorenzo175:
>Do we still need @lboeman's SolarArbiter/solarforecastarbiter-dashboard#270 ?

I think so. This maybe means the download limit is not required, but still probably useful to limit the amount of data plotted for general usability 

> Also can you roughly quantify timing for posterity and reports?

Right now downloading 1 year of 1 minute surfrad data takes ~20s for json and 12s with csv. This PR should bring the json down to around 12s.

wholmgren:
> That does remind me though that we need to implement the splitting of data requests for reports > 1 year and raise the number of days for now in the api used for report generation.

I ran into the 1 year limit when trying to create a report locally. Is the idea that core's api.py would split the request?

Then I hit this when trying to confirm on the dashboard: https://sentry.io/organizations/solararbiter-rx/issues/1888434196/?project=5254377&referrer=alert_email

PR #254, opened 2020-06-04
support for climate zones

### LINKED ISSUE #253: how to handle climate regions
I think we're going to need a standardized way to handle climate regions. While perhaps not required, it might be worthwhile for us to keep mind that the code could be more useful in the long run if it can handle more general regions such as utility/ISO/interconnect boundaries.

Here's one idea:

1. Create database table that tracks known regions. Columns might be `name, kind, specification`. `kind` takes values like `climate` or `iso`. `specification` could be more specific, like `wkt`, or it could let the entry describe itself. Perhaps we'd also want to add some kind of way to control access permissions for arbitrary regions.

2. Create an API function that takes `latitude, longitude` and returns all table rows that include the point. 

3. Create an API function that takes a region name and returns all `datamodel.Site` objects that are included in it. 

4. Add a `regions` attribute to `datamodel.Site`.

I'm thinking of the API functions as things that would recompute the result on demand. But we might want to create additional tables to compute all known associations. That could be tricky to do well with permissions. 

Eventually this information needs to be communicated and searchable in the dashboard. I think the API functions described above meet the requirements for most use cases.

alorenzo175:
closes #253

Adds database tables and procedures to build a mapping between sites and climate zones (stored as native mysql geometry type) since it takes too long to query on demand. Triggers should keep everything up to date with a new zone or site. Right now, there are no permission checks on zones and no way to add new ones through the API.

For the api adds the `/sites/in/<zone>`, /climatezones, `/climatezones/<zone>`, and /climatezones/search endpoints


sfa_api/sites.py
```diff
@@ -242,6 +271,8 @@ def get(self, site_id, *args):
     'sites', 'sites', url_prefix='/sites',
 )
 site_blp.add_url_rule('/', view_func=AllSitesView.as_view('all'))
+site_blp.add_url_rule('/in/<zone>',
+                      view_func=AllSitesInZoneView.as_view('allinzone'))
```
lboeman, line 275: We should add a new URLConverter in `utils/url_converters.py` to avoid a mysql error if the zone name string is >255 characters. Pretty unlikely, but causes an unhandled exception from mysql.

alorenzo175: good call

wholmgren:
awesome that was fast!

PR #256, opened 2020-06-13
add values/gaps endpoints

alorenzo175:
- fix issue in read_write_metadata when the object exists and the user has permissions but the object is not the right type, so should raise mysqlerr
- add values/gaps endpoints to, for example, support finding gaps in persistence forecasts that need to be filled
- add /observations//values/unflagged to find days that are not flagged with a given flag to support https://github.com/SolarArbiter/solarforecastarbiter-core/issues/377
- fix up/add _links to many endpoints

### DIFF
--- a/sfa_api/observations.py
+++ b/sfa_api/observations.py
@@ -19,7 +19,9 @@ from sfa_api.schema import (ObservationValuesSchema,
                             ObservationSchema,
                             ObservationPostSchema,
                             ObservationLinksSchema,
-                            ObservationTimeRangeSchema)
+                            ObservationTimeRangeSchema,
+                            ObservationGapSchema,
+                            ObservationUnflaggedSchema)
 
 
 class AllObservationsView(MethodView):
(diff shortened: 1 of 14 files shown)


sfa_api/observations.py
```diff
@@ -339,6 +342,96 @@ def get(self, observation_id, *args):
...
+            errors = err.errors
+        tz = request.args.get('timezone', 'UTC')
+        flag = request.args.get('flag', None)
+        if tz not in ALLOWED_TIMEZONES:
+            errors['timezone'] = f'Unknown timezone {tz}'
+        if flag is None:
+            errors['flag'] = 'Must provide the flag parameter'
+        else:
+            try:
+                int(flag)
```
lboeman, line 420: I think this should probably have some validation against the allowable flags in core. Although I do like that you could provide a bitmask here because of the `&` operation. There is also a potential data error, which weirdly happens when you exceed the normal 4 byte signed INT maximum and not the unsigned smallint that the column is set to.

alorenzo175: Good catch with the int limit. 

I avoided validating against the core flags because we need the bitmask capability for https://github.com/SolarArbiter/solarforecastarbiter-core/pull/484, and I didn't want to have to possibly deal with different mask versions

PR #265, opened 2020-06-30
add forecast fill to report schema

alorenzo175:
still need tests and final name of field from https://github.com/SolarArbiter/solarforecastarbiter-core/pull/493

### DIFF
--- a/sfa_api/schema.py
+++ b/sfa_api/schema.py
@@ -1143,6 +1143,7 @@ class ReportParameters(ma.Schema):
         title='Forecast Fill Method',
         description=(
             'How to fill missing forecast values before calculating metrics.'),
+        default='drop'
     )
     costs = ma.Nested(
         CostSchema,
--- a/sfa_api/schema.py
+++ b/sfa_api/schema.py
@@ -1096,6 +1096,36 @@ class CostSchema(BaseCostSchema):
     pass
 
 
+class ForecastFillField(ma.Field):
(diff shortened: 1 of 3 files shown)


alorenzo175:
also related https://github.com/SolarArbiter/solarforecastarbiter-dashboard/issues/292

sfa_api/schema.py
```diff
@@ -1116,6 +1139,11 @@ class ReportParameters(ma.Schema):
             " Unlocalized times are assumed to be UTC."),
         validate=TimeLimitValidator()
     )
+    forecast_fill_method = ForecastFillField(
+        title='Forecast Fill Method',
+        description=(
+            'How to fill missing forecast values before calculating metrics.'),
+    )
```
lboeman, line 1146: Adding default and missing values here would be good. Looks like "drop" is the default in core. Could leave out the default to have it be an optional post parameter. Then we wouldn't have to store it, but users would get it in the response.

lboeman: @alorenzo175 Sorry, I should have been more clear, I think having missing here is more important to support existing forecasts, so that the value will be included in responses for reports that are missing this key in the db.

alorenzo175: I think `default` is actually what we want to support existing reports so that the field shows up in the API response even if it is not in the stored json. https://marshmallow.readthedocs.io/en/stable/quickstart.html#specifying-defaults

PR #271, opened 2020-07-21
Limit variables allowed at weather stations

### LINKED ISSUE #241: don't allow sites without modeling_parameters to have power, poa observations
Should not be allowed in the API, nor as an option in the dashboard form.

Sentry Issue: [API-3R](https://sentry.io/organizations/solararbiter-rx/issues/1656781868/?referrer=github_integration)

```
AttributeError: 'Site' object has no attribute 'modeling_parameters'
  File "rq/worker.py", line 886, in perform_job
    rv = job.perform()
  File "rq/job.py", line 664, in perform
    self._result = self._execute()
  File "rq/job.py", line 670, in _execute
    return self.func(*self.args, **self.kwargs)
  File "solarforecastarbiter/validation/tasks.py", line 372, in immediate_observation_validation
    validation_flags = validation_func(observation, value_series)
  File "solarforecastarbiter/validation/tasks.py", line 154, in validate_poa_global
    aoi_func = pvmodel.aoi_func_factory(observation.site.modeling_parameters)
```

lboeman:
closes #241 
This adds a MYSQL procedure to check if a site has modeling parameters by checking for a non-null `tracking_type` column. The procedure is called from python before the subsequent call to `store_obs/fx/cdf` and raises a BadAPIRequestError with an appropriate message. 

@alorenzo175 what do you think of this approach?


alorenzo175:
Hmm, I guess the approach I was thinking about would be to have a function like what you've written, and call that function from within the store_forecast/obs/cdf procedures and abort if appropriate. Otherwise, we could just use read_site and check in python instead of adding a new procedure.

Now that I think about it again, I'm not sure if checking in sql or checking in python is best. Let's go with python for now. We might also want to check more than just tracking type (like capacity). Also need to make sure forecast and aggregate variables match if we don't already

lboeman:
> Hmm, I guess the approach I was thinking about would be to have a function like what you've written, and call that function from within the store_forecast/obs/cdf procedures and abort if appropriate. Otherwise, we could just use read_site and check in python instead of adding a new procedure.
> 
> Now that I think about it again, I'm not sure if checking in sql or checking in python is best. Let's go with python for now. We might also want to check more than just tracking type (like capacity). Also need to make sure forecast and aggregate variables match if we don't already

That's a good point, this procedure is unnecessary. I had originally started working on something in line with what you intended, but I think that keeping the validation in python and the RBAC based control stuff in SQL is a good separation of concerns.

I'll remove the procedure, and check for matching aggregate variable if it doesn't exist.

alorenzo175:
I think the test failures are due to a recent pymysql update. Try restricting to 0.9.3 for datastore/test-requirements.txt and requirements.txt

sfa_api/tests/test_cdf_forecast.py
```diff
@@ -470,3 +470,30 @@ def test_post_forecast_too_large(
             content_type=content_type,
             data=payload, base_url=BASE_URL)
     assert req.status_code == 413
+
+
+@pytest.mark.parametrize('variable', [
+    'ac_power',
+    'dc_power',
+    'poa_global',
+    'availability',
```
alorenzo175, line 479:
```suggestion
    'availability',
    'curtailment'
```

sfa_api/tests/test_forecast.py
```diff
@@ -505,3 +506,36 @@ def test_post_forecast_too_large(
                           'Content-Length': 17*1024*1024},
             data=payload, base_url=BASE_URL)
     assert req.status_code == 413
+
+
+@pytest.mark.parametrize('variable', [
+    'ac_power',
+    'dc_power',
+    'poa_global',
+    'availability',
```
alorenzo175, line 515:
```suggestion
    'availability',
    'curtailment'
```

sfa_api/tests/test_observations.py
```diff
@@ -630,3 +630,17 @@ def test_post_observation_too_large(
                           'Content-Length': 17*1024*1024},
             data=payload, base_url=BASE_URL)
     assert req.status_code == 413
+
+
+@pytest.mark.parametrize('variable', [
+    'ac_power',
+    'dc_power',
+    'poa_global',
+    'availability',
```
alorenzo175, line 639:
```suggestion
    'availability',
    'curtailment'
```

PR #275, opened 2020-09-15
Add deserialization to tz aware field to keep allowable datetimes in

lboeman:
line with pandas.


sfa_api/tests/test_schema.py
```diff
@@ -29,6 +29,9 @@ def test_isodatetime_serialize(inp, out, iso_schema):
 
 @pytest.mark.parametrize('inp,out', [
     ("2019-01-01T00:00:00", dt.datetime(2019, 1, 1, tzinfo=dt.timezone.utc)),
+    ("20190101T000000", dt.datetime(2019, 1, 1, tzinfo=dt.timezone.utc)),
+    ("20190101T00:00:00", dt.datetime(2019, 1, 1, tzinfo=dt.timezone.utc)),
+    ("2019-01-01T000000", dt.datetime(2019, 1, 1, tzinfo=dt.timezone.utc)),
```
alorenzo175, line 34: maybe a few more tests that include a timezone and Z

alorenzo175 (commented):
are the test errors due to pandas > 1.1? do we need to specify a max version in requirements?

sfa_api/tests/test_schema.py
lboeman: > are the test errors due to pandas > 1.1? do we need to specify a max version in requirements?

Yes, I just pushed changes to fix this. Had to set `check_freq=False` to assert_frame_equals calls. I can't find a changelog entry as to why, but the frequency set on the test data breaks things when compared to the other indices.

alorenzo175 (approved):

Issue #147, opened 2019-08-28
report post documentation


wholmgren:
The report POST documentation and examples don't give me enough information to figure out how to format the request. https://api.solarforecastarbiter.org/#tag/Reports/paths/~1reports~1/post

@lboeman can you clarify this doc string?

alorenzo175:
seems like it's taken care of now? https://dev-api.solarforecastarbiter.org/#tag/Reports/paths/~1reports~1/get

alorenzo175:
Looks like the only missing piece is that the filters keys is not well documented. This probably just requires a change in the schema from a list of dicts to a list of a FilterParameter schema

alorenzo175:
if filters need an update in the datamodel, do that first

PR #279, opened 2020-10-06
Improve Report Validation

### LINKED ISSUE #79: missing values on marshmallow deserialize
Many fields in the schema are not required but have no ``missing``or ``default`` values. We should probably set these fields so we know the once place where default values come in. The database storage_interface also won't work as is since the fields are not set at times.

### LINKED ISSUE #193: Improve report validation
There is no validation for requirements such as fx.interval_length < obs.interval_length in the API. This validation is performed in the DataModel and results in a value error thrown in the worker while it attempts to compute the report. From a user perspective, this just results in a failed report with no information about why the failure occurred. 

This issue presents itself directly when running locally with `USE_FAKE_REDIS = TRUE` because the computation is run immediately instead of being handed off to a worker, resulting in a 500.

The dashboard should definitely prevent this in the first place, but it should probably be handled elsewhere as well.  I see there being two ways to handle this:

- Load all of the object pairs and and validate them before storing or enqueueing and return errors.
- Handle errors in workers and store them in the report. This would need to go into core. I'd imagine enabling report templating to handle a failed report and display any errors raised during computation.

@alorenzo175, @wholmgren Do you guys have any thoughts about this? I think this is relatively minor, as the dashboard can help avoid these issues, but it's probably something we should handle eventually.

lboeman:
Trying to tackle a few issues related to Report Validation:
closes #193 - Primary part of this PR.
closes #79 - The remaining point in this issue is to providing missing/defaults to remaining report fields.
#147 - While working with the schema, it's probably worthwhile to improve on the API documentation related to reports.

### DIFF
--- a/sfa_api/conftest.py
+++ b/sfa_api/conftest.py
@@ -105,6 +105,16 @@ VALID_CDF_FORECAST_AGG_JSON.update({
     "constant_values": [10.0, 20.0, 50.0, 80.0, 100.0]
 })
 
+VALID_AGG_JSON = {
+    "name": "Test Aggregate ghi",
+    "variable": "ghi",
+    "interval_label": "ending",
+    "interval_length": 60,
+    "aggregate_type": "sum",
+    "extra_parameters": "extra",
+    "description": "ghi agg",
+    "timezone": "America/Denver"
+}
 
 VALID_OBS_VALUE_JSON = {
     'id': '123e4567-e89b-12d3-a456-426655440000',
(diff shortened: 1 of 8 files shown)


lboeman:
I think this is ready for a review. Should the API reject object pairs with a `forecast_type` of `probabilistic_forecast` that have a reference forecast?

wholmgren:
We're going to add a CRPS skill score soon so I don't think it's worth putting in the effort to reject probabilistic forecasts that have a reference forecast.

lboeman:
> We're going to add a CRPS skill score soon so I don't think it's worth putting in the effort to reject probabilistic forecasts that have a reference forecast.

Are there any further restrictions on probabilistic forecast groups? In this PR, I've only checked matching variable, site or aggregate, interval_length, interval_label and axis.

alorenzo175 (approved):
I feel like there is some special case would still get by this validation, but I can't remember what it was. If it crops up again, we can make a more detailed issue for it.

wholmgren (approved):
no, this is good

PR #280, opened 2020-10-09
Handle permission failure when reading report values

lboeman:
To support viewing a report with only `read` permissions on the report(https://github.com/SolarArbiter/solarforecastarbiter-dashboard/issues/134). Just handles the StorageAuthError that is raised when a user does not have `read_values` and returns an empty list. @alorenzo175 Do you think this needs anything else? Any possible issues with core?

### DIFF
--- a/sfa_api/utils/storage_interface.py
+++ b/sfa_api/utils/storage_interface.py
@@ -1523,7 +1523,10 @@ def read_report(report_id):
     """
     report = _decode_report_parameters(
         _call_procedure_for_single('read_report', report_id))
-    report_values = read_report_values(report_id)
+    try:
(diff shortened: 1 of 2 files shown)


alorenzo175:
I don't think it will cause an issue. Should maybe test w/ a report on dev

lboeman:
I created a new org and test report and shared it with testing@solarforecastarbiter.org to verify this isn't going to fail on dev: https://dev-dashboard.solarforecastarbiter.org/reports/9fa91366-0a76-11eb-92e3-0a580a800215.
OK to merge? Without the dashboard PR, it should just produce empty timeseries/scatter plots.

alorenzo175 (approved):

PR #282, opened 2020-10-12
Adjust deletion procedures and add create_user to admincli

### LINKED ISSUE #122: Able to remove objects from 'applies_to_all' permisison
Users are allowed to run 'remove_object_from_permission' with a permission that has `applies_to_all` set to True.

### LINKED ISSUE #170: delete_user does not remove the user's default role

### LINKED ISSUE #184: add create-user action to admincli
In case the user has made an account but hasn't performed an action to trigger the auto create user and wants to be added to an org.

alorenzo175:
closes #122 closes #170 closes #184

### DIFF
--- a/sfa_api/tests/rbac/test_permissions.py
+++ b/sfa_api/tests/rbac/test_permissions.py
@@ -191,6 +191,21 @@ def test_remove_object_from_permission_object_dne(
     assert objects_on_perm == new_objects_on_perm
 
 
+def test_remove_object_from_permission_object_applies_to_all(
+        api, new_perm,  new_observation):
+    # test that objects list is not altered even though 204 is returned
+    perm_id = new_perm(applies_to_all=True)
+    new_obs = new_observation()
+    # without adding to perm assert observation read allowed
+    get_perm = api.get(f'/permissions/{perm_id}', BASE_URL)
+    perm = get_perm.json
+    objects_on_perm = perm['objects']
+    assert new_obs in objects_on_perm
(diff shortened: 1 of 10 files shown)


sfa_api/tests/rbac/test_permissions.py
```diff
@@ -191,6 +191,21 @@ def test_remove_object_from_permission_object_dne(
...
+    perm_id = new_perm(applies_to_all=True)
+    new_obs = new_observation()
+    # without adding to perm assert observation read allowed
+    get_perm = api.get(f'/permissions/{perm_id}', BASE_URL)
+    perm = get_perm.json
+    objects_on_perm = perm['objects']
+    assert new_obs in objects_on_perm
+    remove_object = api.delete(f'/permissions/{perm_id}/objects/{new_obs}',
+                               BASE_URL)
+    assert remove_object.status_code == 404
```
lboeman, line 206: Maybe add another get of the permission here to assert that permission hasn't been removed.

lboeman (approved):
Loogs good!

sfa_api/tests/rbac/test_permissions.py
alorenzo175: I agree, but it seems like the mysql error it raises cancels the transaction so that the observation and permission are gone after the 404. At least manually checking w/o using transactions, it works as expected

lboeman: Ah, yeah. forgot about that hickup :+1:

PR #285, opened 2020-10-14
Update API docs

### LINKED ISSUE #47: JSON missing/null value parsing
If we're expecting users to serialize missing or null values into JSON as the javascript `null` object, Flask will parse it into None and pandas into a NaN.
If we need to be more permissive with missing data in JSON, we'll have to find/replace accepted NaN values.

### LINKED ISSUE #59: Set Proper API responses
All endpoints should respond with:
- a 200 code when successful
- 401: unauthorized when a user is not logged in (this may need to be merged with 404 depending on if we require signup for public data.
- 400 when a logged in user has permissions to perform an action but the request is invalid for some other reason. This usually produces a meaningful error message.
- 404 User lacks permissions or the resource does not exist. This is so that users cannot discover resources they do not have access to.

### LINKED ISSUE #251: document units of probabilistic forecast values
What needs to be in %, what needs to be in physical units? is anything in fractional units?

### LINKED ISSUE #281: Add report status options to docs
Report status doesn't have options/descriptions.

alorenzo175:
closes #59 closes #281 closes #251 closes #47

### DIFF
--- a/sfa_api/aggregates.py
+++ b/sfa_api/aggregates.py
@@ -347,6 +347,8 @@ class AggregateForecasts(MethodView):
                     $ref: '#/components/schemas/ForecastMetadata'
           401:
             $ref: '#/components/responses/401-Unauthorized'
+          404:
+             $ref: '#/components/responses/404-NotFound'
         """
         storage = get_storage()
         forecasts = storage.list_forecasts(aggregate_id=aggregate_id)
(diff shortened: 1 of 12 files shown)


lboeman (approved):
Looks good to me. I made a PR to your branch to add request params and the csv nan example to the CDFForecastValues get endpoint.

PR #287, opened 2020-10-15
Aggregate values

### LINKED ISSUE #219: Fix aggregate when observation effective_until is set and requested range is after that
right now, no data for that observation would be returned and then the aggregate computation function chokes

### LINKED ISSUE #255: Index Error when observation deleted and in aggregate
at least in the case when the aggregate has only a single observation. Comes from core.compute_aggregate checking if the observation is valid. Reproduce by running dev api, creating aggregate with a single observation. Then delete the observation and try and access the aggregate

lboeman:
Creates and passes an index for computing aggregate values to the `solarforecastarbiter.utils.compute_aggregate` function updated in https://github.com/SolarArbiter/solarforecastarbiter-core/pull/590

The `_make aggregate_index` function was enforcing a rule that aggregate interval lengths are divisors of one day, So I've added a validator for the aggregate schema to enforce this. 
closes #255 
closes #219

### DIFF
--- a/sfa_api/aggregates.py
+++ b/sfa_api/aggregates.py
@@ -184,23 +184,31 @@ class AggregateValuesView(MethodView):
         aggregate = storage.read_aggregate(aggregate_id)
         indv_obs = storage.read_aggregate_values(aggregate_id, start, end)
 
-        aggregate_interval = f"{aggregate['interval_length']}min"
+        interval_length = f"{aggregate['interval_length']}min"
+        interval_label = aggregate['interval_label']
+        timezone = aggregate['timezone']
 
-        index_start = start.tz_convert(
-            aggregate['timezone']
-        ).ceil(aggregate_interval)
+        if interval_label == 'ending':
+            index_start = start.tz_convert(
+                timezone
+            ).ceil(interval_length)
+            index_end = end.tz_convert(timezone)
+        else:
+            index_start = start.tz_convert(timezone)
+            index_end = end.tz_convert(
+                timezone
+            ).floor(interval_length)
 
         request_index = pd.date_range(
             index_start,
-            end.tz_convert(aggregate['timezone']),
-            freq=aggregate_interval,
+            index_end,
+            freq=interval_length,
         )
 
         # compute agg
         try:
             values = compute_aggregate(
-                indv_obs, aggregate_interval,
-                aggregate['interval_label'], aggregate['timezone'],
+                indv_obs, interval_length, interval_label, timezone,
                 aggregate['aggregate_type'], aggregate['observations'],
                 request_index)
         except (KeyError, ValueError) as err:
--- a/sfa_api/aggregates.py
+++ b/sfa_api/aggregates.py
@@ -190,15 +190,15 @@ class AggregateValuesView(MethodView):
             end.tz_convert(aggregate['timezone']),
             freq=f"{aggregate['interval_length']}min"
         )
-        request_index = request_index[request_index > start]
+        request_index = request_index[request_index >= start]
 
         # compute agg
         try:
             values = compute_aggregate(
-                request_index,
                 indv_obs, f"{aggregate['interval_length']}min",
                 aggregate['interval_label'], aggregate['timezone'],
-                aggregate['aggregate_type'], aggregate['observations'])
+                aggregate['aggregate_type'], aggregate['observations'],
+                request_index)
         except (KeyError, ValueError) as err:
             raise BaseAPIException(422, values=str(err))
         accepts = request.accept_mimetypes.best_match(['application/json',
--- a/sfa_api/aggregates.py
+++ b/sfa_api/aggregates.py
@@ -184,18 +184,22 @@ class AggregateValuesView(MethodView):
         aggregate = storage.read_aggregate(aggregate_id)
         indv_obs = storage.read_aggregate_values(aggregate_id, start, end)
 
-        start_midnight = start.tz_convert(aggregate['timezone']).floor('D')
+        aggregate_interval = f"{aggregate['interval_length']}min"
+
+        index_start = start.tz_convert(
+            aggregate['timezone']
+        ).ceil(aggregate_interval)
+
         request_index = pd.date_range(
-            start_midnight,
+            index_start,
             end.tz_convert(aggregate['timezone']),
-            freq=f"{aggregate['interval_length']}min"
+            freq=aggregate_interval,
         )
-        request_index = request_index[request_index >= start]
 
         # compute agg
         try:
             values = compute_aggregate(
-                indv_obs, f"{aggregate['interval_length']}min",
+                indv_obs, aggregate_interval,
                 aggregate['interval_label'], aggregate['timezone'],
                 aggregate['aggregate_type'], aggregate['observations'],
                 request_index)
--- a/sfa_api/aggregates.py
+++ b/sfa_api/aggregates.py
@@ -189,19 +189,15 @@ class AggregateValuesView(MethodView):
         timezone = aggregate['timezone']
 
         if interval_label == 'ending':
-            index_start = start.tz_convert(
-                timezone
-            ).ceil(interval_length)
-            index_end = end.tz_convert(timezone)
+            index_start = start.ceil(interval_length)
+            index_end = end.ceil(interval_length)
         else:
-            index_start = start.tz_convert(timezone)
-            index_end = end.tz_convert(
-                timezone
-            ).floor(interval_length)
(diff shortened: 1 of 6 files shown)


sfa_api/aggregates.py
```diff
@@ -181,12 +182,26 @@ def get(self, aggregate_id, *args):
         storage = get_storage()
         aggregate = storage.read_aggregate(aggregate_id)
         indv_obs = storage.read_aggregate_values(aggregate_id, start, end)
+
+        aggregate_interval = f"{aggregate['interval_length']}min"
+
+        index_start = start.tz_convert(
+            aggregate['timezone']
+        ).ceil(aggregate_interval)
```
alorenzo175, line 190: hmm, shouldn't this take into account the interval label like core.utils._make_aggregate_index?

alorenzo175 (commented):
should add a test explicitly for the situation described in #255

lboeman:
> should add a test explicitly for the situation described in #255

Done, though I may have misunderstood this error, and neglected it in the core PR. The KeyError was already being caught here and results in an error message on the dashboard. Should this error only occur if the deleted observation is valid at the given period?

sfa_api/aggregates.py
lboeman: I realized the aggregate endpoint behaved such that the `start` and `end` query parameters were defining the bounds of the observation data to aggregate, rather than the aggregate datapoints to return. So, I've adjusted the start and end parameter before reading, to include all the observation data expected in the first/last interval. 
I'm also always using `start.ceil(interval_length)` to find the index start, and `end.floor(interval_length)` to find the index end. This makes the start and end inclusive, consistent with the other endpoints.

alorenzo175:
> > should add a test explicitly for the situation described in #255
> 
> Done, though I may have misunderstood this error, and neglected it in the core PR. The KeyError was already being caught here and results in an error message on the dashboard. Should this error only occur if the deleted observation is valid at the given period?

I'm not sure. Really great issue... I do think the core change would take care of it as far as I understand the issue

sfa_api/tests/test_aggregates.py
```diff
@@ -378,7 +378,17 @@ def test_get_aggregate_values_limited_effective(api, aggregate_id, startend):
                   headers={'Accept': 'application/json'},
                   base_url=BASE_URL)
     assert res.status_code == 200
-    assert not math.isnan(res.json['values'][-1]['value'])
```
alorenzo175, line 381: what caused this change?

sfa_api/tests/test_aggregates.py
```diff
@@ -378,7 +378,17 @@ def test_get_aggregate_values_limited_effective(api, aggregate_id, startend):
...
+    assert res.json['values'][-1]['value'] is None
+
+
+def test_get_aggregate_values_outside_effective(api, aggregate_id):
```
alorenzo175, line 384: is this effectively the same as the test on ln 333?

sfa_api/aggregates.py
```diff
@@ -181,13 +182,39 @@ def get(self, aggregate_id, *args):
...
+
+        interval_length = f"{aggregate['interval_length']}min"
+        interval_label = aggregate['interval_label']
+        timezone = aggregate['timezone']
+
+        # Create a timedelta to add/substract from end/start to get data
+        # outside of start/end when aggregating
+        interval_offset = pd.Timedelta(interval_length) - pd.Timedelta('1ns')
+
+        index_start = start.ceil(interval_length)
```
alorenzo175, line 194: I don't understand the switch from floor/ceil based on interval label. Like if I requested a time at the middle of  an interval with interval_label beginning, that first interval is basically thrown out. But if i request the same time w/ interval_label ending, we'll go back and get all the data for the interval.

alorenzo175 (commented):
I'm confused about a few things, but I might just not be thinking straight

Maybe the tests have changed, or I just didn't see it before, but I don't see a new test for #219 i.e. set effective_until and make a request for data after that. I guess that should be a 422 since there are no valid observations?

sfa_api/tests/test_aggregates.py
lboeman: Ah, you're right. I added this after altering core. This error was raised with the "Cannot aggregate without missing ids..." message before. I'll consolidate to a single test. I also thought that this was

sfa_api/aggregates.py
lboeman: I don't think I totally follow what you're saying here, but I'll try to explain my logic.

In my thinking any request should only return timestamps for the aggregate inside of [start, end]. Taking ceiling of start gets us the next timestamp after start(or start), and floor gets us the last timestamp before end(or end). Then we can create an index including all of the timestamps inside [start, end].

Then based on the interval label, we adjust the range of values we read for the observations forward/backward so that we're including all of the data for the first/final interval.

e.g. in the test `test_aggregate_values_interval_label_offset_request` with  `start=2019-04-14T13:30Z&end=2019-04-14T14:30Z`, an index with just '2019-04-14T14:00Z' is created. When interval label is `beginning` the end parameter for `read_aggregate_values` is adjusted to '2019-04-14T14:59:59...Z' so that we can aggregate with the whole hour of data when core does its resampling. Then the data before that point is discarded.

sfa_api/tests/test_aggregates.py
lboeman: I've updated this with some comments and more assertions.
The data returned by the api includes all of the expected datapoints between 2019-01-01T00:00Z and 2020-01-01T00:00Z filled with NaNs for missing observation data, so there's just a chunk of real values between 2019-04-14T07:00Z and 2019-04-17T0655Z.

sfa_api/aggregates.py
alorenzo175: I'm thinking of it more as that a request should aggregate data between [start, end]. So if `start=2019-04-14T13:30Z`, data at that time should be included. If `interval_label=beginning` and `interval_length=60min`, as it is now, only data from 2019-04-14T14:00Z to 2019-04-14T14:59:59.999999999Z  would be aggregated. This PR would change the behaviour from what is currently returned (although that doesn't always have enough data to aggregate properly).

An interesting corner case is `start=2019-04-14T13:30Z`, `end=2019-04-14T13:59Z`, `interval_label=beginning`, 'interval_length=60min`. Here `index_start=2019-04-14T14:00Z` and `index_end=2019-04-14T13:00Z`, so no data would be returned (or maybe an error in compute aggregate). I think that would confuse users.

lboeman: I see what you're saying now. I see the usefulness of being able to request the range of data to aggregate, but I find a couple of things confusing about it, at least when considered with the behavior of the other types of objects.

The meaning of the data returned for a request for values from an observation and aggregate with the same interval length and interval label, both representing the interval mean, would be different. Unless the user performs the offset themselves. 
e.g. data for 12:00Z-12:00Z vs data for 12:00Z-12:59Z timestamps should be the same for either request, but the values returned in the aggregate request change if it includes a shorter interval length observation.

Stemming from that, It seems like when comparing a forecast to an aggregate, it possibly introduces inaccuracies at either end of the report period, being that the forecast will be made for the full intervals. This may have a near negligible effect over the span of the report period though.

Also if an aggregate includes observations of differing interval_length, say a 5 and 60, assuming the intervals align with each hour, a request like `start=2019-04-14T13:30Z`, `end=2019-04-14T13:59Z`, `interval_label=beginning`, 'interval_length=60min` would not include the data for the larger interval and I think should be invalid.

I also realized that we're not considering observation interval labels when we aggregate data, which has it's own handful of issues, but probably a stretch for this PR.

My intention was that the corner case you propose return no data, but currently raises an error from `compute_aggregate`.

Either way we go here, I'll add some tests to reinforce the behavior pertaining to the issues discussed here with some notes on the decisions made.

alorenzo175: Ah sorry, what I meant was that you do adjust index_start/start and index_end/end as appropriate for the aggregate *not* that you only get data from start to end and then aggregate. So in the example with `start=2019-04-14T13:30Z` and `end=2019-04-14T13:59Z', you would request data from 13:00Z to 13:59:59.999999999Z and the output would be a single value at 13:00Z. 

Thus, the start time of 13:30Z is included in the aggregate, but we still extend so that it properly represents the aggregate interval_length.

lboeman: Ah, okay. Yes, I agree with that.

lboeman:
@alorenzo175 I think I've brought this in line with your suggestions. Could you give it another review?

alorenzo175 (approved):

PR #289, opened 2020-10-19
Make uncertainty optional

### LINKED ISSUE #234: make uncertainty optional
With the way we're using uncertainty in core and in the dashboard report form, I think it's reasonable to relax the requirement that users supply an uncertainty.

alorenzo175:
closes #234


alorenzo175:
aggregate pr should handle the test error

lboeman (approved):

PR #297, opened 2020-10-28
Convert None to NaN in timeseries before calling compute_aggregate

### LINKED ISSUE #296: Non-Numeric NaN values cause aggregate resampling to crash.
Data read from the database contains `None` as the nan value in the dataframe. The Core `compute_aggregation` function chokes when pandas does not know what to do with this Non-numeric type. 
Sentry Issue: [API-5K](https://sentry.io/organizations/solararbiter-rx/issues/1985464536/?referrer=github_integration)

```
DataError: No numeric types to aggregate
(7 additional frame(s) were not displayed)
...
  File "flask/views.py", line 163, in dispatch_request
    return meth(*args, **kwargs)
  File "sfa_api/aggregates.py", line 189, in get
    aggregate['aggregate_type'], aggregate['observations'])
  File "solarforecastarbiter/utils.py", line 156, in compute_aggregate
    new_val = resampler['value'].mean().reindex(new_index)
  File "pandas/core/groupby/groupby.py", line 1225, in mean
    "mean", alt=lambda x, axis: Series(x).mean(**kwargs), **kwargs
  File "pandas/core/groupby/groupby.py", line 907, in _cython_agg_general
    raise DataError("No numeric types to aggregate")
```

lboeman:
closes #296 
Converts None values to NaNs to avoid pandas resampling choking with "No numeric types to aggregate" error in core. I've done this here because the core api module handles this conversion for users, so I think this is something the api should take care of. This could also be handled at the storage interface level if there's any concern the issue might crop up elsewhere.

### DIFF
--- a/sfa_api/aggregates.py
+++ b/sfa_api/aggregates.py
@@ -5,6 +5,7 @@ from functools import partial
 from flask import Blueprint, request, jsonify, make_response, url_for
 from flask.views import MethodView
 from marshmallow import ValidationError
+import numpy as np
 import pandas as pd
 from solarforecastarbiter.utils import compute_aggregate
 
@@ -25,7 +26,7 @@ from sfa_api.schema import (AggregateSchema,
 class AllAggregatesView(MethodView):
     def get(self, *args):
         """
-        ---
+ nan)       ---
         summary: List aggregates.
         description: List all aggregates that the user has access to.
         tags:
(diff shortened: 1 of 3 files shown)


alorenzo175:
Yeah, I think this should be handled at the storage interface level, along with forecast values and observation values. I think the best way to go about it might be to set the `dtype=float` when constructing the dataframes

alorenzo175 (approved):
add the `as_type` to the read_latest_*_values  too

PR #299, opened 2020-11-03
Remove report values not present in a raw_report update

### LINKED ISSUE #298: Remove old report processed values
Recently, we've seen that the report json for some of the SURFRAD sites is over 200MB in size of almost entirely processed data values. It turns out when recomputing a report (which has happend 223 times so far for the Bondville GHI report), the processed values are added to the database, but the old values are not removed. So fetching the report gets all 1784 of these processed values instead of the expected 8.

Since the values and raw report are posted separately, perhaps the best way to go about this is to require a list of processed values to keep be included with the raw report on POST. Then, as the raw report is being stored all other processed values are deleted.

alorenzo175:
closes #298

### DIFF
--- a/datastore/migrations/0053_remove_old_report_obj.up.sql
+++ b/datastore/migrations/0053_remove_old_report_obj.up.sql
@@ -6,7 +6,8 @@ BEGIN
     DECLARE binid BINARY(16);
     DECLARE allowed BOOLEAN DEFAULT FALSE;
     SET binid = UUID_TO_BIN(strid, 1);
-    SET allowed = can_user_perform_action(auth0id, binid, 'update');
+    SET allowed = (can_user_perform_action(auth0id, binid, 'update') AND
+        can_user_perform_action(auth0id, binid, 'write_values'));
     IF allowed THEN
         DELETE FROM arbiter_data.report_values WHERE report_id = binid AND id NOT IN (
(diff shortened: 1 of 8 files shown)


datastore/migrations/0053_remove_old_report_obj.up.sql
```diff
@@ -0,0 +1,24 @@
+DROP PROCEDURE store_raw_report;
+CREATE DEFINER = 'insert_objects'@'localhost' PROCEDURE store_raw_report(
+IN auth0id VARCHAR(31), IN strid CHAR(36), IN new_raw_report JSON, IN keep_report_values JSON)
+COMMENT 'Update metrics field with json and raw_report with binary data'
+BEGIN
+    DECLARE binid BINARY(16);
+    DECLARE allowed BOOLEAN DEFAULT FALSE;
+    SET binid = UUID_TO_BIN(strid, 1);
+    SET allowed = can_user_perform_action(auth0id, binid, 'update');
```
lboeman, line 9: We might want to check for 'write_values' here as well to protect against the unlikely event that a user without it posts only a raw report and tanks the values.

lboeman (approved):
Looks good. Minor concern about the permissions for posting a raw_report.

datastore/migrations/0053_remove_old_report_obj.up.sql
alorenzo175: good call

PR #306, opened 2020-11-16
Add api auth examples and post examples

### LINKED ISSUE #301: add data post w/ curl example
curl -H "Authorization: Bearer $ACCESS_TOKEN" https://dev-api.solarforecastarbiter.org/forecasts/single/f8dd49fa-23e2-48a0-862b-ba0af6dec276/values --data-binary @test.csv  -H 'content-type: text/csv'

### LINKED ISSUE #303: authorization doc suggestions
A few things to add to the authorization section since it confused at least one user.

The `curl/jq` example should make it clear that this is just one way of doing it. We could even start with plain HTTP and then provide the curl/jq example.

It might also be useful if we at least linked to `request_cli_access_token` function. Maybe even inline it.

https://github.com/SolarArbiter/solarforecastarbiter-core/blob/4e5808e5719bf5e60f1a79f5c92cf0f7576c4827/solarforecastarbiter/io/api.py#L35-L61

Next, we should provide more concrete guidance on how frequently to request a new token. Maybe change/move the sentence that starts with "A JWT will expire after a set period...". If I remember correctly, another user was rate limited by auth0 when trying to make a lot of requests, so we should include something about that too.

The sentence on access control could link to the relevant API sections and/or https://solarforecastarbiter.org/documentation/dashboard/administration/

alorenzo175:
and require that tokens be signed via RS256

closes #301 closes #303

### DIFF
--- a/datastore/docker-compose.yml
+++ b/datastore/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3'
 services:
   mysql:
-    image: percona:8.0-centos
+    image: percona:8.0.20-11-centos
     restart: always
     environment:
       MYSQL_ROOT_PASSWORD: 'testpassword'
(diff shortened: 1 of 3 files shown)


wholmgren (approved):
thanks this helps a lot

PR #307, opened 2020-11-18
test \ufeff in csv and add ref aggregate permissions

alorenzo175:


sfa_api/utils/request_handling.py
```diff
@@ -128,7 +129,8 @@ def parse_json(json_str):
         values key cannot be parsed into a DataFrame.
     """
     try:
-        json_dict = json.loads(json_str)
+        json_dict = json.loads(''.join(s for s in json_str
+                                       if s in string.printable))
```
lboeman, line 133: Might be able to use `json_str.lstrip('\ufeff')` here, but this might defend against other unknown encoding problems.

lboeman (approved):

PR #313, opened 2021-07-08
add report_parameters.timezone

wholmgren:
supports https://github.com/SolarArbiter/solarforecastarbiter-core/pull/681

### DIFF
--- a/sfa_api/schema.py
+++ b/sfa_api/schema.py
@@ -1455,6 +1455,16 @@ class ReportParameters(ma.Schema):
         description="List of categories with which to group metrics.",
         required=True
     )
+    timezone = ma.String(
+        title="Timezone",
+        description=(
+            "IANA Timezone. Data will be converted to this timezone before "
+            "statistics are calculated. If not provided, timezone will be "
+            "inferred from sites."
+        ),
+        required=False,
+        validate=TimezoneValidator()
+    )
 
     @validates_schema
     def validate_cost(self, data, **kwargs):
--- a/sfa_api/schema.py
+++ b/sfa_api/schema.py
@@ -1462,7 +1462,8 @@ class ReportParameters(ma.Schema):
             "statistics are calculated. If not provided, timezone will be "
             "inferred from sites."
         ),
-        required=False,
+        default=None,
+        missing=None,
         validate=TimezoneValidator()
     )
 
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,6 @@ venv.bak/
 # vim
 *.swo
 *.swp
+
+# vscode
+.vscode/
--- a/sfa_api/conftest.py
+++ b/sfa_api/conftest.py
@@ -222,7 +222,8 @@ def report_post_json():
                     'observation': '123e4567-e89b-12d3-a456-426655440000',
                     'forecast': '11c20780-76ae-4b11-bef1-7a75bdc784e3'
                 }
-            ]
+            ],
+            'timezone': None
         }
     }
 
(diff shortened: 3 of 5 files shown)


sfa_api/schema.py
```diff
@@ -1455,6 +1455,16 @@ class ReportParameters(ma.Schema):
...
     )
+    timezone = ma.String(
+        title="Timezone",
+        description=(
+            "IANA Timezone. Data will be converted to this timezone before "
+            "statistics are calculated. If not provided, timezone will be "
+            "inferred from sites."
+        ),
+        required=False,
+        validate=TimezoneValidator()
```
lboeman, line 1466: We might need to add  `missing=None` here to fill in the value if users don't specify the key in `report_parameters`.

lboeman (commented):
I think this is all that we need in terms of for the actual api code. A timezone parameter will need to be added to the fixture here: https://github.com/SolarArbiter/solarforecastarbiter-api/blob/32138986499ca0b0b9ab8c1001ccb5f09c20c534/sfa_api/conftest.py#L208 and probably need to add a migration to insert a null timezone into the parameters of existing reports.

sfa_api/schema.py
wholmgren: I'm not sure what the best approach is on this. The existing reports don't have a `timezone` parameter within their `report_parameters`, so it would mark a break between the expectations for the old and the new reports. In practice, the dashboard could either post `None` (or `null`?) or it could just not include the field if it's not set. 

To be clear, the changes in core do not currently require that `report_parameters` contains `timezone`.

lboeman: Looking again, that's actually the how we're handing other optional fields so okay to leave as is, and saves work updating the db. We might need to add  `allow_none=True` if we want to allow for posting null values.

wholmgren: > We might need to add allow_none=True if we want to allow for posting null values.

I guess I'm neutral on allowing `"timezone": null` vs just having people leave it out. Do we have a standard? Would that work with the validator or would we need a new one?

lboeman: urgh. I was off base with my previous comment. The current behavior for other optional report parameters is to provide a missing and default value as with costs here: https://github.com/SolarArbiter/solarforecastarbiter-api/blob/32138986499ca0b0b9ab8c1001ccb5f09c20c534/sfa_api/schema.py#L1406-L1412. So we end up with those default values in the API response. 
The missing value provides a value at deserialization and the default value  provides it at serialization, so we wouldn't have to mess with the db and we'd store None for newly created reports.
If the value is none, validation will not be performed so the existing validator with suffice.

wholmgren: Ah, so add `default=None` and `missing=None`? And then I also remove `required=False`?

lboeman: Yep!

wholmgren:
> A timezone parameter will need to be added to the fixture here: 

I believe this is no longer necessary/desired given the default/missing parameters. Adding it caused a few tests to fail, while removing it made the tests pass again. Also the other optional parameter, forecast fill method, isn't included in that fixture.

Looks to me like the test python37 failure is unrelated. And not sure why codecov is complaining since it says the diff is fully covered.

lboeman (approved):

wholmgren:
@alorenzo175 any concerns with this or the related core PR?

alorenzo175:
Nope, no concerns from me. I would just make sure to make a report now on dev (with no timezone specified) and make sure it still works after merging this and core (without wiping the dev db). This will hopefully catch any issues that could affect the prod reports.

wholmgren:
thanks!

